PR #4, opened 2015-07-17
Sync

priyavadan:


ginty:
Thanks a lot @priyavadan, if you can keep these coming periodically it would be much appreciated!

priyavadan:
Sounds good!

PR #12, opened 2015-09-04
Sync to Internal version v2.5.0.pre149

priyavadan:
- Changes made to internal version to support the transition have been omiited.

### DIFF
--- a/lib/origen/specs/spec.rb
+++ b/lib/origen/specs/spec.rb
@@ -113,6 +113,70 @@ module Origen
         super
       end
 
+      # Returns the trace_matrix name.  The Trace Matrix Name is composed of
+      # * @name
+      # * @type
+      # * @subtype
+      # * @mode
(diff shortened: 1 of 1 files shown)


ginty:
Thanks!

priyavadan:
np. I did not add extra documentation, left a note with David so he can add it if required.

Also, tried to omit stuff that was internal release only. If I missed something, please let me know and i will revisit this sync.

ginty:
No that was perfect thanks. Honestly some of this stuff should maybe be
removed, is trace matrix really a universal concept for example? We can
revisit later though.

priyavadan:
I think the RTM (requirement trace matrix) is a pretty standard/universal way to check if requirements are met and help create the document but I guess that the method can be made a little more generic? I'll let David answer this in detail, whenever he joins :)

ginty:
Ok cool, that's good to know.

PR #13, opened 2015-09-05
Modified initialize method in revision_control/git

priyavadan:
- The initialize method will return true if
 - .git exists
  - it is NOT the Initial Commit.

Any reason why the check requires that the remote URL matches https://github.com/Origen-SDK/origen.git?  

From my understanding, the remote origin by default will point to the forked repository URL. 
I do not think this check is required. Not sure if I am missing something obvious.

### DIFF
--- a/lib/origen/revision_control/git.rb
+++ b/lib/origen/revision_control/git.rb
@@ -229,7 +229,7 @@ module Origen
 
       def initialized?
         File.exist?("#{local}/.git") &&
-      #    git('remote -v', verbose: false).any? { |r| r =~ /#{remote_without_protocol_and_user}/ || r =~ /#{remote_without_protocol_and_user.to_s.gsub(':', "\/")}/ } &&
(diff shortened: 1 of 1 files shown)


priyavadan:
if it's okay, I'll remove the commented out code so that coverage is not reduced.

ginty:
The intention here is not only to check that the given local dir is a git repo, but also that it is an initialized version of the remote repo that you want.

I suppose in the mismatch case it should really raise an error rather than return false.

Generally though, this API can be hit outwith the context of an Origen app workspace, you can do things like:

git = Origen::RevisionControl::Git.new(local: "/some/path/to/existing/dir", remote: "git://blah/blah")

git.initialized?   # => Should probably error if the local is a git repo to a different remote

So I would like to preserve this feature, and probably improve it to raise an error.

priyavadan:
Will re-open this request later

PR #14, opened 2015-09-12
Initial Checkin of VagrantFile

priyavadan:
To start the preconfigured virtual machine you will need the following installed
 - Vagrant https://www.vagrantup.com/
 - Virtualbox https://www.virtualbox.org/

 - To start type 'vagrant up' from the root of the directory where you cloned origen
    - Vagrant server now has all dependencies and origen v0.4.1 installed by default.
    - The origen repository you just cloned will be mounted at '/vagrant' inside the virtual machine.
    - type 'origen -v' to install the required gems automatically and report the origen version.


priyavadan:
please don't merge yet, I will probably update the documentation pages in the next few days and explain the process in detail.

priyavadan:
please go ahead and try the vagrant setup with this branch and let me know if you have any recommendations.

ginty:
Good stuff, got it to run although the origen new command didn't work right due to some problem with .files not being pulled down correctly from app_generators. Didn't investigate though.

I think the image should be called origen/0.4.1 or similar rather than priyavadan/.

So I'm in two minds about whether this is a good way to run Origen on windows or not, I think on balance though I prefer the native install.

Pros for this:

* Origen installation details can be completely managed by us
* We don't really need to care about the windows platform in Origen code anymore

Cons for it:

* I don't think the installation is any easier than it would be to install Origen natively - e.g. I had to install vagrant, ssh, and probably a good terminal client
* The download footprint is huge, much bigger than for a native install. Normally this wouldn't be much of a problem but it could be in the scenario of folk wanting to try it out during a conference.
* You can't use native windows editors and git GUIs for example, which are things that a windows user would likely want to do.

In terms of ITC, I do think it would be better spending the next few weeks getting the native windows solution working and well documented.

priyavadan:
Stephen,

Thanks for the feedback. My intent was to show that using the pre-configured box will definitely be the easiest way to play around with Origen after the conference and if someone has the bandwidth to download it, well good on them :+1: 

Funny you mention it, I am currently documenting the native windows install method that could be used while in the conference.

- Trying to make it as user friendly as possible and from the command line (optional but including installing vagrant and virtual-box)

- There will be about 7 commands the user will need to type to have a fully functional ruby install on windows (ruby 2.1.6 for now, will upgrade to 2.2.2 when some of the issues are fixed by the developers)

 - The issue is Rubygems 2.4 is broken on windows, so need a bit of a trick to get it to work correctly.

 - Good news is that if the commands are followed as shown (or will be shown in the document) it will work flawlessly
 
- It might require downloading the 7 packages from the internet but these are small packages. 

- I think this might be acceptable and a nice compromise since even if we provide a complete batch file package, users will still need to download that from the internet during the conference.

priyavadan:
also, the origen testers plugin needs an update to function correctly on windows, will submit a pull request soon with the required updates.

priyavadan:
To answer some more questions that II missed

-  I agree, once we have the core team members try out the image and provide feedback, we will release v1 of the Origen-SDK/origen image 

- the files stored in /Vagrant are accessible in windows, so user can edit the files in their favorite windows editor and run the commands in the Linux VM

- The user would need git installed on windows (easily available installer and will be a one command install when I finish the windows guide). Once git is installed, Vagrant ssh will connect to the VM directly from the console(or command prompt) so no terminal client should be required unless the user prefers a different client.

priyavadan:
Have updated the vagrantcloud image to Origen-SDK/origen

ginty:
@priyavadan, do you still want this?

priyavadan:
na, don't think we are ever going to use Vagrant/Docker for Origen, I will close this PR. Can re-open if we ever change our minds.

PR #21, opened 2015-09-28
Regression Fix + Exit code updates

priyavadan:


ginty:
Hey, issue here is that coverage is based on running specs and examples within a single thread, so it can't exit when printing the result of the first one

priyavadan:
ah ic, okay, hmm will need to rethink

PR #29, opened 2016-02-13
Improve exhibits.rb, fix issues in specs.rb, and expand the specs.rb self-test

welguisz:
Find some instances where knowing the override information is needed for exhibits.  Added that to this code.

Improved Continuous Integration Environment for specs.rb.  This find several issues in specs.rb.  Coverage for specs.rb should be 90%+ now.


ginty:
Nice, thanks!

PR #32, opened 2016-02-29
Here is an initial attempt to model pins that are not directly associated with the SoC.  For use in defining pins that need to be a part of the ATE test program (e.g. DIB utility channels for relays, sma observation points, socketID, etc.), but are not top-level DUT pins.

aaronburg:


ginty:
Hi @aaronburg,

Looks good.

Only thing I am wondering, is if we should rename these to be virtual pins or something similar? I do feel slightly uneasy about introducing a very tester-specific concept like a utility bit control into Origen core.

Could we change the API to be 'add_virtual_pin', and then you could use that within your application to model utility bit control, while still allowing other apps/domains to use it for something else?

aaronburg:
That seems reasonable.  Really my only requirement is that this new category of pins has it's own API access so that it does not appear in the $dut.pins list - so a virtual pin would be ok.  Are you suggesting just changing the nomenclature of the existing API or something more involved?

ginty:
Yeah, potentially just find "utility" and replace with "virtual".

aaronburg:
Done and checked in...

ginty:
Nice udpate @aaronburg, thanks!

Bonus points for a follow up PR to add a few comments on this new feature to: http://origen-sdk.org/origen/guides/models/pins/

aaronburg:
Done.  Pull request made.

Regards,
Aaron

PR #34, opened 2016-03-09
New Class for Specs Documentation

welguisz:
This class allows for documentation to control how the map output and the topic output would look.  This is just the base class and individual companies and people can add on to make it specific to their strategy.

### DIFF
--- a/lib/origen/specs.rb
+++ b/lib/origen/specs.rb
@@ -10,10 +10,11 @@ module Origen
     autoload :Version_History, 'origen/specs/version_history.rb'
     autoload :Creation_Info, 'origen/specs/creation_info.rb'
     autoload :Spec_Features, 'origen/specs/spec_features.rb'
+    autoload :Documentation, 'origen/specs/documentation.rb'
     require 'origen/specs/checkers'
     include Checkers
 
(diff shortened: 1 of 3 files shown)


priyavadan:
@welguisz this needs a lint fix :)

welguisz:
Wait until my next commit.  I noticed that I left out some spec tests.

welguisz:
This looks good to go.  Can I get a new release of Origen in the next 24 hours so I can continue on some of my projects.  Thanks!

lib/origen/specs.rb
```diff
@@ -324,6 +332,53 @@ def doc_resources(options = {})
...
+      }.update(options)
+      return nil if @_documentation.nil?
+      return nil if @_documentation.empty?
+      doc_found = Hash.new do |h, k|
+        h[k] = Hash.new do |hh, kk|
+          hh[kk] = Hash.new do |hhh, kkk|
+            hhh[kkk] = Hash.new do |hhhh, kkkk|
+              hhhh[kkkk] = Hash.new do |hhhhh, kkkkk|
+                hhhhh[kkkkk] = Hash.new do |hhhhhh, kkkkkk|
+                  hhhhhh[kkkkkk] = {}
```
ginty: Yikes!
Don't need to change it right now, but this trick should help with this kind of thing:
https://github.com/franzejr/best-ruby/blob/master/tricks/autovivification.md

PR #36, opened 2016-03-11
added capability to support automated regression testing

priyavadan:
- when workflow is git (i.e: echo latest | origen regression... will check against latest and not require user input)
- this should help with regressions using continous integration systems


ginty:
LGTM

priyavadan:
@ginty thanks!

PR #37, opened 2016-03-11
Fix masking

chrisnappi:
Unfortunately found another mask issue.  This fixes it.

### DIFF
--- a/lib/origen/pins/pin_collection.rb
+++ b/lib/origen/pins/pin_collection.rb
@@ -326,9 +326,10 @@ module Origen
 
       def assert(value, options = {})
         each_with_index do |pin, i|
-          if !value.respond_to?('is_to_be_read?') || value[size - i - 1].is_to_be_read?
-            value = value.data if value.respond_to?('data')
+          if !value.respond_to?('data')
             pin.assert(value[size - i - 1], options)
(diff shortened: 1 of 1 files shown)


chrisnappi:
FYI - with the old code it would not over-mask.

lib/origen/pins/pin_collection.rb
```diff
@@ -326,9 +326,10 @@ def cycle
 
       def assert(value, options = {})
         each_with_index do |pin, i|
-          if !value.respond_to?('is_to_be_read?') || value[size - i - 1].is_to_be_read?
-            value = value.data if value.respond_to?('data')
+          if !value.respond_to?('data')
             pin.assert(value[size - i - 1], options)
+          elsif value[size - i - 1].is_to_be_read?
+            pin.assert(value.data[size - i - 1], options)
```
ginty: Though they both work, value[size - i - 1].data would probably be better here, or just what you had before, value[size - i - 1] to pass the bit to pin.assert.

The way it is currently written converts the whole register to a value each time around the loop, which seems a bit wasteful. In most cases it probably doesn't matter, but pin/vector level APIs can potentially be called thousands of times within a given pattern and it all adds up.

chrisnappi:
Will resubmit with suggested change.

PR #38, opened 2016-03-14
Have the ability to checkin to a branch that is not master

welguisz:
I was working on some code and no matter what I did, I could not get Origen to check in the file on a different branch than master.  When I did the following things, I was able to get Origen to check in the file on a different branch:

* Added 'options' to the _checkout unless options[:initial]_ to allow for version to be passed
* In my code, added version option to the checkin command.

When I did that I got the ability to check in files on a different branch.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -179,6 +179,3 @@ DEPENDENCIES
   origen_doc_helpers (>= 0.2.0)
   origen_testers (~> 0)
   stackprof (~> 0)
-
-BUNDLED WITH
-   1.11.2
--- a/lib/origen/revision_control/git.rb
+++ b/lib/origen/revision_control/git.rb
@@ -88,7 +88,7 @@ module Origen
           git("checkout #{local_rev} -- #{paths.join(' ')}", check_errors: false)
           # Then proceed with checking them in as latest
         else
-          checkout unless options[:initial]
+          checkout options unless options[:initial]
         end
         cmd = 'add'
         if options[:unmanaged]


welguisz:
Looking through the error logs, both say:

```
The plugins API
  returns an enhanced array containing all plugin app instances
[DEPRECATED] 11.116[0.091]   || Origen.app.plugins_manager and Origen.app.current_plugin are deprecated, use Origen.app.plugins instead
[DEPRECATED] 11.116[0.000]   || Called by /home/travis/build/Origen-SDK/origen/spec/plugin_api_spec.rb:17
[DEPRECATED] 11.116[0.000]   || Origen.current_plugin.name is deprecated, use Origen.app.plugins.current.name instead
[DEPRECATED] 11.116[0.000]   || Called by /home/travis/build/Origen-SDK/origen/spec/plugin_api_spec.rb:17
  can set and get the current plugin (FAILED - 1)
  can temporarily disable the current plugin (FAILED - 2)
  methods can be overridden by the application
  methods can be added by the application
  Origen.root references within a plugin mean the top-level app root
  Origen.root! references within a plugin mean the plugin root
  Origen.root! references within a top level app are equivalent to Origen.root
  Origen.app references within a plugin mean the top-level app
  Origen.app! references within a plugin mean the plugin app
  Origen.app! references within a top level app are equivalent to Origen.app
  Shared commands work
```

No idea why that is failing when I just changed one line in git.rb

ginty:
Hi @welguisz, sorry I broke the master branch yesterday, so forget about those test fails.

I would like to understand the background to this one a bit more, can you give some examples of what you were trying that didn't work before?

Thanks.

welguisz:
So I have an Origen app where the end user wants the ability to select a branch to read/write to.  On the command line, they type _origen --branch featureA_.  They want to read their branch, have the App run, and make changes back to the branch *featureA*.  

When I was trying to fix this within my app, the checkout worked fine.  When I checked them in, they went to the master which would cause problems.

With the checkout comment, there is options[:version] which selects the branch and it works.  When I tried with the current options for checkin, I was unable to select the branch.

#### Example
Assumption:
* You have a repository that is mainly data that multiple projects query

##### Scenarios
###### Bug-Fixes/Add new Feature
* The source has an issue.
* Contributor create a branch (e.g. bug-fix/description OR feature/new-feature)
* Contributor commits and push their fixes to their forked repository and their branch
* Contributor runs _origen --branch feature/new-feature_.
* Contributor see the results and decide if they need to do more fixes or ready for pull-request
* Contributor likes the results and makes a pull request to the Main Repository for *develop* or *their_project*

###### Branches for their projects
* Project moves from development to production
* They need to branch their data so new data does not overwrite their data set.
* An issue is find and they need traceability.
 * If there is just one branch called *master*, traceability is lost
 * If there is a branch called *projectA*, changes can be made to that branch without crosstalk

I can give you a more concrete example via e-mail if needed.

ginty:
So if you call checkout with no arguments it is definitely supposed to checkout the current branch unless you specify otherwise, this is the code to do that:

https://github.com/welguisz/origen/blob/master/lib/origen/revision_control/git.rb#L29

Could you possibly investigate what current_branch returns in your case and why it doesn't work?

welguisz:
I think we can remove this fix.  It was user issue.  In my options I had [:initial] set to true which caused master to be used.

ginty:
Ok cool, thanks for following up.

PR #45, opened 2016-04-06
Added capability to find specs using :symbol attribute

aaronburg:
Added capability to the 'show_specs' method to allow the user to find all the specs wth a given spec symbol, rather than by spec name.

### DIFF
--- a/spec/specs_spec.rb
+++ b/spec/specs_spec.rb
@@ -293,6 +293,11 @@ describe "Origen Specs Module" do
     $dut.has_spec?(:tnikhov2).should == true
     $dut.has_spec?(/tnikhov/).should == true
   end
+
+  it "finding specs via :symbol option works" do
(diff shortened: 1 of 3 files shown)


spec/specs_spec.rb
```diff
@@ -293,6 +293,11 @@ def get_true_hash_size(hash, obj_class)
     $dut.has_spec?(:tnikhov2).should == true
     $dut.has_spec?(/tnikhov/).should == true
   end
+
+  it "finding specs via :symbol option works" do
+    $dut.has_spec?(nil, symbol: :gvdd).should == true
```
ginty: I think this would be more idiomatic Origen:

~~~ruby
$dut.has_spec?(:gvdd, symbol: true)
~~~

What do you think?

aaronburg: @ginty
Yep, I like it.  I’ll make the change.

PR #49, opened 2016-05-03
Remove limit checker on Nokogiri::XML::NodeSet

welguisz:
Added Nokogiri::XML::NodeSet as a class to ignore when evaluating limits.


welguisz:
@ginty or @priyavadan Possible to get a release out for this when you have some time?

Issue #52, opened 2016-06-01
Need to compact system gem specs during origen gem initialization


priyavadan:
If the system gem specs array has a nil element, it will results in an exception which will cause issues running Origen.

priyavadan:
Fixed in v0.7.13

Issue #53, opened 2016-06-01
Force overwrite of the system gem if it already exists


priyavadan:
FileUtils.cp currently does not overwrite an existing gem but throws an exception. Either force overwrite or handle the exception and continue.

ginty:
Not sure that it should force an overwrite here, I think it should just do nothing if the user's gem directory already has it.

priyavadan:
@ginty yup, my bad, it should look if file exists and skip. PR #54 has the updates.

priyavadan:
Fixed in v0.7.13

PR #54, opened 2016-06-02
Bugfix/regression

priyavadan:
Fixes Issue #52 and #53


bin/origen
```diff
@@ -134,16 +136,37 @@ if origen_root && File.exist?(ENV['BUNDLE_GEMFILE']) && Origen.site_config.gem_m
                   (spec.executables || []).each do |bin|
                     FileUtils.cp("#{spec.bin_dir}/#{bin}", "#{local_gem_dir}/bin")
                   end
-                  FileUtils.cp(spec.cache_file, "#{local_gem_dir}/cache")
+                  if File.exist(spec.cache_file)
```
ginty: Don't think it needs to be so noisy when things are going OK, I would remove all the puts statements

priyavadan: updated

ginty:
LGTM, thanks!

priyavadan:
Thanks! origen_core 0.7.13 has been released.

PR #55, opened 2016-06-13
Minor update to regression manager

priyavadan:
- added capability to provide the reference workspace path and version number.
- fixed minor bug while setting up the origen workspace. For some reason, bundle install is also needed along with origen -v in a newly built workspace.


ginty:
LGTM

PR #57, opened 2016-08-08
Adds support for MSB0 register bit order (RFC 0001)

ginty:
This PR implements the bit order feature discussed in [RFC 0001](https://github.com/Origen-SDK/rfcs/pull/1).

All Origen models now have a bit_order attribute, which is inherited from their parent if not otherwise set, and defaults to :lsb0 which is the name given to the bit order that has been used until now.
By setting a model's @bit_order attribute to :msb0, then all registers that are descendants of that block will have their bit order reversed.
This attribute can also be set at individual register level to override the inherited bit order.
### Implications of Using MSB0 Bit Order

The bit_order attribute is really describing the physical layout of the register, rather than its logical behavior, the main effect then is that in the console the register will be displayed backwards.

The existing methods shift_out_left and shift_out_right are referring to the physical view of the register as it appears in the console. Since the physical view is reversed it means that the data order returned by these methods will be reversed. These are the methods that are commonly used by Origen serial interface drivers and therefore by assigning a register as :msb0 its data value will be reversed in a JTAG or similar serial protocol.

I had some concern in the RFC discussion that this might mean that any references in IP-level test code like my_reg[5] or my_reg.my_bits[5..3] would break if instantiated in a parent SoC model that switched the bit order, however that is not the case. Such references are to logical rather than physical bit location, and therefore data bit 5 is always data bit 5 regardless of the physical orientation and such code will remain valid.
### New BitCollection Methods

This PR also adds two new methods for serially working with register data:
- shift_out - A logical rather than a physical shift, this will stream the data LSB first, and will be the same regardless of the register's bit order.
- reverse_shift_out - As above but MSB first.

It also adds with_index versions of all shift methods:
- shift_out_left_with_index
- shift_out_right_with_index
- shift_out_with_index
- reverse_shift_out_with_index
### What's Not Supported

There was some discussion in the RFC about adding the ability to override the bit order at bit level, however that has not been supported since upon implementation is doesn't really seem to make much sense.
An MSB0 register where all bits have LSB0 bit order is just the same as an LSB0 register. Having the case where only a subset of bits have different bit order does not seem to be something anyone would ever do, and it would add significant complication to support it.

Also MSB0 bit order has not been tested on registers with split bits (where members of the same bus signal are not adjacent within the register). If required this can be added in a future update


ginty:
@ra2727dude, please give this a try and feedback if it is to your liking.
Add it to your Gemfile like this:

~~~ruby
gem 'origen', :git => 'https://github.com/Origen-SDK/origen.git', :branch => 'msb0_reg_bit_order'
~~~

ginty:
@chrisnappi, @chrishume, @priyavadan please review

ginty:
@ra2727dude, I'm going to go ahead and merge this soon if there is no further feedback

ra2727dude:
Hi Stephen, I tested the functionality and it meets all expectations.  Please merge at your convenience.

PR #58, opened 2016-08-21
Store Log Messages into a Hash for debug purposes

welguisz:
In an application, I need to be able to parse the log messages being sent.  At the end of my app, I need to go through and categorize Error, Warning, and Success messages depending on where they occur.

For example, if I am importing something and it passes, I want to be able to write:
`
   Origen.log.success('Block A', :import)
`

If it fails, I want to write:
`
   Origen.log.error('Block B', :import)
`

At the end of my app, I want to output the errors depending on the section.
~~~~
Origen.log.msg_hash[:error].each do |section, arr|
   arr.each do |a|
    p "#{section.to_s.upcase}::: #{a}"
  end
end
~~~~

### DIFF
--- a/lib/origen/log.rb
+++ b/lib/origen/log.rb
@@ -12,11 +12,25 @@ module Origen
     require 'log4r'
     require 'log4r/outputter/fileoutputter'
 
+    attr_accessor :msg_hash
+
     LEVELS = [:normal, :verbose, :silent]
 
     def initialize
       @log_time_0 = @t0 = Time.new
       self.level = :normal
+      @msg_hash = init_msg_hash
+    end
+
+    def init_msg_hash
+      msg_types = [:info, :warn, :error, :deprecate, :debug, :success]
+      msg_hash = {}
+      msg_types.each do |m|
+        msg_hash[m] = Hash.new do |h, k|
+          h[k] = []
+        end
+      end
+      msg_hash
     end
 
     def console_only?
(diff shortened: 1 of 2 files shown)


lib/origen/log.rb
```diff
@@ -74,50 +88,56 @@ def level
       @level
     end
 
-    def debug(string = '')
+    def debug(string = '', msg_type = nil)
```
ginty: Does this work ok if the string and msg_type and both missing?
Could you add a test for that to make sure?

i.e. this is allowed by the existing API and is used to print a blank line in the log:

~~~ruby
Origen.log.debug
Origen.log.debug "This is important so I added some padding"
Origen.log.debug
~~~

welguisz:
@ginty I added a check to make sure that Origen.log.debug (and all of the other logging methods) can handle both msg and msg_type being nil.

It works as expected.

ginty:
Thanks @welguisz.
I think this case will probably not work right as it stands, do you want to support it?

~~~ruby
Origen.log.debug :section1
Origen.log.debug "This is important so I added some padding", :section1
Origen.log.debug :section1
~~~

welguisz:
@ginty ... Good catch. It will not support it. It should be able to support the padding with msg not set. I will update the tests and check to see that it works as expected.

ginty:
Included in 0.7.26 - http://origen-sdk.org/origen/release_notes/#v0_7_26

PR #59, opened 2016-09-09
Add Other Pins to Pin Group

welguisz:
In prepping for a new project, I needed the ability to handle Other Pins.  Other pins include the following types:

*  No-Connects
*  Reserved
*  DePopulated Pins

Used Virtual Pin as a base for the Other Pins and copied its spec tests to ensure that the code works as expected.


ginty:
Looks good, note that one of your new tests is failing and will need to be fixed before merging

welguisz:
@ginty spec test is passing now.

PR #75, opened 2017-03-17
Pin.clock enhancement to handle non-50%-duty-cycle

rlaj:
Enhance pin clock feature to handle cases where 50% duty cycle isn't best fit.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source 'https://rubygems.org'
+source 'http://rubygems.org'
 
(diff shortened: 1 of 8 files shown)


ginty:
Very cool, any chance this can include a doc update?

Just a section in here on how to generate free-running clocks would be fine: http://origen-sdk.org/origen/guides/pattern/common/

rlaj:
Sure, I'll add it.

PR #74, opened 2017-02-10
Added better pin meta data API and some power pin attributes

ginty:
@welguisz could you give this a review pls?

It gives every pin a default meta data hash:

~~~ruby
add_pin :pinx

pins(:pinx).meta   # => {}
~~~

And allows it to be overridden by function:

~~~ruby
add_pin :pin2, meta: { vol: 0.4, voh: 0.6 }   # Define default meta data

pin(:pin2).add_function :tdo
pin(:pin2).add_function :nvm_done, meta: { voh: 0.7 }  # Override some of the meta data

pin(:tdo).meta         # => { vol: 0.4, voh: 0.6 }
pin(:nvm_done).meta    # => { vol: 0.4, voh: 0.7 }
~~~

The above merging of default and function-specific hashes will actually be done for any FUNCTION_SCOPED_ATTRIBUTE which is defined as a Hash.

However if the attribute is not a Hash then it will skip the merging, so it should play nice with any legacy apps where the meta data is not assigned to a hash, see test for this:

https://github.com/Origen-SDK/origen/compare/pins_meta_data?expand=1#diff-423cda76f858a0dd741ff2cb11cb9a12R598

Finally it also adds some additional attributes for power pins:

~~~ruby
add_power_pin :vddio, voltage: 3, current_limit: 50.mA, meta: { min_voltage: 1.5 }

power_pin(:vddio).voltage             # => 3
power_pin(:vddio).current_limit       # => 50E-03
power_pin(:vddio).meta[:min_voltage]  # => 1.5
~~~

PR #76, opened 2017-03-17
Timing api

ginty:
This PR adds an API to define pin timing.
Initially, this is being used by the origen_sim plugin to create simulation waveforms, but the intention is that it will also be used by OrigenTesters in the future to generate ATE timing sheets.

This branch also contains the updates in #74, I will ensure that that is approved and merged before merging this one.
It also contains a very minor update to allow the reg.sync API to be supported in simulations.

Here now is a description of the new API, once this is reviewed and agreed I will copy this to a guide page before merging.

### Pin Timing API - Defining Timesets

A wave object defines the waveform that should be applied to a pin, and each pin will have two wave objects assigned to it - one for drive cycles and one for cycles where the pin is being read/compared.

A timeset object is used to collect the wave objects for all pins. Multiple timesets can be defined and therefore the waveforms being applied to the pins can be changed by changing the timeset.
The API to change the timeset is the existing one provided by OrigenTesters:

~~~ruby
tester.set_timeset('func', 100)
~~~

The period of the timeset can be referenced in the wave definitions to make the waveforms keep the same shape with different period settings.

Here is the most basic timeset definition, this will give all pins a default waveform which drives for the whole period on drive cycles, and strobes at 50% of compare cycles:

~~~ruby
# Simple definition, all pins have default waves
add_timeset :t1
~~~

Here is a more complex definition, which changes the default compare for all pins to be at 25% of the period, and which adds a unique drive wave to the :tck pin to create a clock:

~~~ruby
# Complex definition, defines an alternative default compare wave and specific timing for :tck
timeset :func do |t|
  t.wave do |w|
    w.compare :data, at: "period / 4"
  end

  t.wave :tck do |w|
     w.drive :data, at: 0
     w.drive 0, at: 25
     w.dont_care at: "period - 10"  # Just to show that dont_care can be used
  end
end
~~~

Pin groups can also be referenced in the timeset definition, here to add common drive timing to all pins in :gpio, and with a special compare waveform for :gpio5 only:

~~~ruby
# Another timeset to show the wave assignment to pin groups
timeset :t2 do |t|
  t.wave :gpio5 do |w|
    w.compare :data, at: 100
  end

  t.wave :gpio do |w|
    w.drive :data, at: 200
  end
end
~~~

The API currently only supports edge compares, i.e. window compares cannot be defined.

### Consuming Timeset Data

The currently selected timeset can be retrieved via the following methods which are aliases:

~~~ruby
dut.timeset               # => instance of Origen::Pins::Timing::Timeset
dut.current_timeset
~~~

Using the plural returns a hash containing all timesets, thereby allowing access to specific timesets regardless of the current selection:

~~~ruby
dut.timesets          # => Hash
dut.timesets[:func]   # => instance of Origen::Pins::Timing::Timeset
# This will also work:
dut.timeset(:func)   
~~~

Each timeset stores the waves in two arrays, one for drive waves and another for the compare waves:

~~~ruby
dut.timeset(:func).drive_waves    # => Array containing instances of Origen::Pins::Timing::Wave
dut.timeset(:func).compare_waves
~~~

Each wave object has an events array, and the pins assigned to it can also be retrieved:

~~~ruby
drive_wave = dut.timeset(:func).drive_waves.first
compare_wave = dut.timeset(:func).compare_waves.first

drive_wave.events    # => [[0, :data], [50, 0], [75, :x]]
drive_wave.pins      # => Array of Origen::Pins::Pin instances

compare_wave.events  # => [["period / 2", :data]]
~~~

The method evaluated_events can be used to get the events with the formulas evaluated based on the current period:

~~~ruby
compare_wave.events            # => [["period / 2", :data]]
compare_wave.evaluated_events  # => [[50, :data]]
~~~

The waves assigned to a given pin by the current timeset can also be accessed via the pin API:

~~~ruby
dut.pin(:tms).compare_wave.events[0]   # => ["period / 2", :data]
~~~


ginty:
@pderouen can you please give this a look over, it would be good to integrate this with OrigenLink in future and replace the custom timing API you have there

pderouen:
@ginty, based on the description in the comments it seems like it would be easy to translate this timing information into something understood by OrigenLink.  OrigenLink just needs to know what order to use when operating on the pins (defined by the order provided when defining the timeset).  For a simple return low/high clock, it just needs to know when to operate on the pins in relation to the clock pin (provides 3 buckets):

timing value:              0   1   2
clock waveform:      ___/***\___

Inside the timing buckets, the order you provide the pins to link.pintiming determines when they are operated on relative to each other.

Let me know when you're ready to try it out.

pderouen:
@ginty, is this api intended to be used for rendering timeset sheets in a test program as well, or just for the simulator and link?  Not being able to set compare and drive timing for a pin would be problematic for rendering a timeset sheet in IG-XL.

ginty:
@pderouen, yes in future I expect OrigenTesters to be able to render such a definition to a timing sheet.
You can set compare and drive timing for a pin, you just have to define them separately in different waves:

~~~ruby
timeset :func do |t|
  t.wave :swdio do |w|
    w.compare :data, at: "period / 4"
  end

  t.wave :swdio do |w|
     w.drive :data, at: 0
  end
end
~~~

You can't define a wave to compare and drive within the same cycle though, do you think that's necessary?

pderouen:
@ginty, no I don't think it's necessary.  I just misunderstood the error message.

ginty:
I wonder if changing the API to make it more explicit which one you are defining would help:

~~~ruby
timeset :func do |t|
  t.drive_wave :swdio do |w|
    w.compare :data, at: "period / 4"
  end

  t.compare_wave :swdio do |w|
     w.drive :data, at: 0
  end
end
~~~

pderouen:
That seems more intuitive to me.

pderouen:
@ginty, after executing this:
~~~ruby
    tester.set_timeset('tp0', 40)
    dut.timeset :tp0 do |t|
      t.wave :tck do |w|
        w.drive :data, at: 20
      end
      t.wave :tdi, :tdo, :tms do |w|
        w.drive :data, at: 5
      end
      t.wave do |w|
        w.compare :data, at: 35
      end
    end
~~~
These attributes return nil:
~~~ruby
dut.current_timeset_period
dut.current_timeset
~~~
Are these attributes supposed to be set by the set_timeset method?

ginty:
Yeah it looks like OrigenTesters will need an update, or an extension to it added to this plugin.

Currently, it seems I am setting it via the OrigenSim tester driver:

https://github.com/Origen-SDK/origen_sim/blob/master/lib/origen_sim/tester.rb#L36

pderouen:
Does the dut not have access to read the timeset name and period from the tester object?  If it does you could avoid the testers update.

ginty:
You could, but I think the dependency is the wrong way around in that case.
Essentially, this is Origen core, which should be a standalone, top-level entity without dependencies on plugins like Testers.

So with this change we are saying that Origen core now provides support for tracking the current timeset, and this is the API to use it:

~~~ruby
dut.current_timeset_period = 100
dut.current_timeset = "func"
~~~

Now, OrigenTesters already provides a legacy API which does sort of the same thing, and which can be considered syntactic sugar on top of this lower level API:

~~~ruby
tester.set_timeset("func", 100)
~~~

That is obviously widely used and we want to maintain it, however it is the responsibility of OrigenTesters to do the linkage to the core platform API and not the other way around.

Issue #19, opened 2015-09-25
Remotes feature should be updated to use abstract revision control API


ginty:
The remotes feature should be updated to use the abstract RC API under the hood:
http://origen-sdk.org/origen/guides/misc/revisioncontrol/

This will update the remotes feature to work with any RC tool supported by Origen.

Also please update the remotes guide to use this API to reference the path to the imported data:
http://origen-sdk.org/origen/release_notes/#v0_5_6

rlaj:
Implemented in PR #78. Closing.

PR #78, opened 2017-03-20
Update remotes feature to use abstract RC API, Issue #19

rlaj:


ginty:
Don't think it needs to be changed, but just to comment that rather than maintaining local files to keep track of things, you can use the session API for this kind of thing: http://origen-sdk.org/origen//guides/misc/session/

PR #84, opened 2017-05-11
Moved nokogiri from 1.6.7.2 to 1.7.2

welguisz:
This covers the issue seen in [Origen Issue #83](https://github.com/Origen-SDK/origen/issues/83).

If we move Origen to this version, I think we would need to update the version from 0.7.47 to 0.8.0 since we will be upgrading form Ruby 2.1.5 to Ruby 2.3.


ginty:
I think I would try only changing the spec line to:

~~~ruby
 spec.add_runtime_dependency "nokogiri", ">=1.6.7.2"
~~~

Also see the discussion in #83

ginty:
Origen has been updated to Nokogiri 1.7.2 as of v0.11.0.

PR #87, opened 2017-06-07
Bye GCC, hello Regexp core extensions plus a fix for the 'command_spec' test

info-rchitect:
* Added in a Regexp core extension library
* Fixed a 'commands' spec test that did not take into account 'x64-mingw32' platforms
* Removed GCC dependency


spec/commands_spec.rb
```diff
@@ -7,13 +7,13 @@
     output.should include "Application: #{Origen.app.version}"
     output.should include "Origen: #{Origen.version}"
 
-    if RUBY_PLATFORM == "i386-mingw32" then
+    if ['i386-mingw32','x64-mingw32'].include? RUBY_PLATFORM
```
ginty: Do you mind changing this to:

~~~ruby
if Origen.os.windows?
~~~

ginty:
Thanks for the update @info-rchitect, can you run "origen lint" to clean up the lint error that is causing the build to fail please

ginty:
Nice update @info-rchitect, thanks!

ginty:
Included in 0.13.0

PR #88, opened 2017-06-07
Added support for a global Origen session, from the user's perspectiv…

coreyeng:
Added support for a global Origen session, from the user's perspective, that is stored at ~/.origen/.session. This behaves the same as the application session but is accessible between Origen workspaces.

Added method Origen.home to point to ~/.origen. Actual implementation uses Ruby's Dir.home.

Added some cases to check if the 'app' is the Origen module and update files and paths accordingly.

Added a 'private' accessor that will make the session file only read/writable by the user. Permissions: 600. Default is 664. This is available for any session, global or application. Added .private? method as well to check if this is set.

Changed around the return strings if the 'app' is Origen. Reads as 'Origen's Global Session'.

Updated session.md.erb accordingly.

### DIFF
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -683,6 +683,17 @@ unless defined? RGen::ORIGENTRANSITION
         @current_command
       end
 
+      # Provides a global Origen session stored at ~/.origen/session (Origen.home)
+      def session(reload = false)
+        @session = nil if reload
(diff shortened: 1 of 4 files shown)


priyavadan:
looks good to me!

ginty:
Nice update, and a great PR, thanks!

ginty:
Included in 0.13.0

PR #89, opened 2017-06-13
Added support for NXDI during LSF initialization 

priyavadan:
Added support for NXDI during LSF initialization as Queue names are different.

Will add a different method later to ask for user input and maintain it in an Origen session.

### DIFF
--- a/lib/origen/application/lsf.rb
+++ b/lib/origen/application/lsf.rb
@@ -24,11 +24,21 @@ module Origen
           @group = nil
           @project = 'msg.te'
           @debug = false
+          require 'timeout'
           puts 'Please enter one of the values from the proposed list'
-          @queue = get_text(suggested_values: Origen.lsf.queuenames, single: true)
-          # There is no command yet to find all the available resource values, so providing a list here of
-          # 2 values that are most commonly used.
-          @resource = get_text(suggested_values: 'linux, rhel6', single: true)
+          begin
+            # Default Timeout for input is 15 seconds
+            Timeout.timeout 15 do
+              @queue = get_text(suggested_values: Origen.lsf.queuenames, single: true, default: 'short')
+              # There is no command yet to find all the available resource values, so providing a list here of
+              # 2 values that are most commonly used.
+              @resource = get_text(suggested_values: 'linux, rhel6', single: true, default: 'linux')
+            end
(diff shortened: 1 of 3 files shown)


lib/origen/application/lsf.rb
```diff
@@ -22,9 +23,31 @@ class Configuration
...
+          # The resource is currently assigned as either linux or rhel6 as well dependent on the queue assignment
+          `bqueues > /tmp/queuenames.log`
+          file = File.open('/tmp/queuenames.log')
+          columns = []
+          file.each_line do |line|
+            columns << line.split(' ')[0, 11]
+          end
+          queuenames = []
+          # rubocop:disable Style/For: Prefer each over for
+          for i in 1..(columns.size - 1)
```
ginty: Should listen to rubocop here, should be:

~~~ruby
(1..(columns.size - 1)).each do |i|
  #...
~~~

priyavadan:
@ginty I understand that we need to think of making the lsf configuration something that can be change in an environment. Will give it some thought.

If you/anyone has an idea, I would love to hear it.

ginty:
Hi @priyavadan, sorry it took so long to get back to you on this.

Fundamentally the site_config is Origen's mechanism for customizing Origen on a per company/environment basis and LSF configuration certainly seems to fit into that category. Basically, therefore I think we should be looking at a solution here which expands the site_config to add LSF variables.

To future proof it I think we should add support for the following site config variables (and probably more for whatever LSF customization variables we have today):

* default_lsf_queue

Not really supported today, but in future would like to support being able to identify patterns as long and short execution, which may mean they would have a preference for different queues, if defined, if not would fall back to the default:

* short_lsf_queue  
* long_lsf_queue

Using the site_config gives you the ability to control these at the site, application or user level.

For the LSF params though, some of them you may want to adjust on a per target level or with some other dynamic selection performed at runtime. Such things are not supported by the site config, so you may want to think about that here too.

Potentially moving the defaults to the site config system, but still having the application config able to override them like you can today would solve that.

info-rchitect:
@ginty @priyavadan can we close this PR?

priyavadan:
@info-rchitect not yet, this still needs work but other priorities are keeping me busy.

priyavadan:
A record gets broken today! This PR will be initiated again once more work is done on this feature :)

PR #92, opened 2017-06-27
Bug fix - @max_address_reg_size with single-reg sub-block

chrishume:
Initialized @max_address_reg_size to prevent nil case for single-register sub-block.


priyavadan:
looks good to me!

Issue #93, opened 2017-06-28
Ambiguous program file error is really a bug?


info-rchitect:
Hi,

As you can see below, I do not have 2 versions of the 'ws1.rb' program flow file.  Bug?

thx

~~~shell
peologin01:stp $ origen p ws1.rb
[INFO]       0.012[0.012]    || **********************************************************************
The following matches were found:
/home/username/origen/stp/program/ws1.rb
/home/username/origen/stp/program/ws1.rb

COMPLETE CALL STACK
-------------------
Ambiguous file ws1.rb
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:183:in `all_matches'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:123:in `clean_path_to'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:73:in `block in resolve_files'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:71:in `each'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:71:in `resolve_files'
/home/username/.origen/gems/ruby/2.3.0/gems/origen_testers-0.9.6/lib/origen_testers/origen_ext/generator.rb:14:in `generate_prog
ram'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/application/runner.rb:81:in `block (2 levels) in launch'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/application/runner.rb:140:in `record_invocation'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/application/runner.rb:75:in `block in launch'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/file_handler.rb:392:in `preserve_state'
/home/username/.origen/gems/ruby/2.3.0/gems/origen-0.20.1/lib/origen/application/runner.rb:18:in `launch'
/home/username/.origen/gems/ruby/2.3.0/gems/origen_testers-0.9.6/lib/origen_testers/origen_ext/application/runner.rb:21:in `laun          ch'
[...]

PR #95, opened 2017-06-29
Fixed a bug where the file handler was erroring when it found the sam…

info-rchitect:
…e file twice with Dir.glob.  Was due to the matches.flatten.uniq call being inside a conditional.  Bug only occurred if user does not specify the program dir when running origen -p

I looked to add in a spec test but wasn't sure where to put it.


ginty:
Many thanks @info-rchitect!

Fix for #93

ginty:
PR's that add any additional test coverage are always very welcome.

I didn't insist on it in this case as it is in an area that is likely to be subject to very little churn in the future.

PR #98, opened 2017-07-01
Added in 'has_key?' and 'rm_session_file' methods to the session store

info-rchitect:
convenient way to check if a key already exists.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -86,6 +86,8 @@ GEM
     net-ldap (0.16.0)
     nokogiri (1.7.2)
       mini_portile2 (~> 2.1.0)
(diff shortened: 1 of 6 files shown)


ginty:
Looks fine, need to fix the build error though (most likely a lint error) before we can merge

info-rchitect:
is there something I need to fix?

ginty:
Looks like it, the build is still failing, you should be able to click on that and click through to a log to find out what the problem is

info-rchitect:
OK  I cannot see anything wrong with that file.  it shows I have added some methods and that is it.  origen specs passes with flying colors.  wtf?

PR #104, opened 2017-07-15
Additional session store methods and basic spec tests

info-rchitect:


info-rchitect:
I don't see how the coverage could go down so much, i didn't touch any of the files coveralls is complaining about.  I added a new spec file database_spec.rb so the coverage should go up.

ginty:
Do we need to have the keys and user_keys methods?

Should we just have the keys method which does what the user_keys method does?

info-rchitect:
well the users_keys gets rid of the keys that the user did not make, in essence the overhead keys.  we could make it an option for the keys method:

~~~ruby
def keys(options = {})
  options = {
    user_created_only: false
  }.update(options)
  if options[:user_created_only]
    store.keys.reject { |k| k.to_s.match(/refresh_interval_in_minutes/) }
  else
    store.keys
  end
end
~~~

Thoughts?

ginty:
Understood, the question is really why would a user ever want to know about the internal use/overhead keys?

info-rchitect:
I don't think they would, they just want their keys.  So you mean you want this:

~~~ruby
def keys
    store.keys.reject { |k| k.to_s.match(/refresh_interval_in_minutes/) }
end
~~~

ginty:
Yep, though something like this should be easier to extend in future if we need to screen more keys:

~~~ruby
def keys
  store.keys - [:refresh_interval_in_minutes]
end
~~~

ginty:
Is this comment about not being able to delete items from the session no longer valid?

https://github.com/Origen-SDK/origen/blob/master/templates/web/guides/misc/session.md.erb#L54

If so, could you update it accordingly?
Thanks!

info-rchitect:
The CI build failure was only on one version of Ruby and it has nothing to do with the PR.  Can we approve this?

image

info-rchitect:
thanks very much!

ginty:
Included in 0.20.3

PR #105, opened 2017-07-19
Global commands

coreyeng:
We've been using Origen here to generate patterns with great success. So much success in fact, that we started building other, non-pattern-generation related utilities using Origen. These take advantage of several items that Origen gives us for free, such as; revision control, web documentation, email notification, and LDAP authentication services. This had inevitably led to us developing plugins in Origen for more generalized aspects (basically, just scripting) but that don't require any of the 'application' stuff Origen has. That is, these plugins aren't tied to any application and could care less if that application even exists. These plugins have shared commands that could be run with or without an application.

Currently, the only way to do this is to add a custom site config (possibly several, depending on how your environment is structured) that disables Bundler and allows one to manager their installation individually. However, we lose a huge asset (Bundler) when we do this. Additionally, this requires more user invention than should be necessary (making custom site configs, then debugging those configs, etc.). Alternatively, you could just create a dummy application and load your Gemfile with the tools required, but then you are required to invoke Origen from that application in order to do anything. This is just an inconvenience, but an inconvenience that I think could be fixed.

This pull request's purpose is allow user customization of their global installation AFTER first getting the general configuration from the site config. This allows them to customize their global Origen version and dependencies, including those that contain globally available commands.

The new flow in bin/origen is to first look if Origen is invoked in an application and run lbin/origen if so. If not, Origen will look at ~/.origen and see if a lbin/origen exists there. If so, that will be run, essentially allowing Origen to start up as if it were within an application, only at ~/.origen instead. If this doesn't exists, Origen will start up from the site config. However, the first startup of the site config with be akin to running Origen for the first time in an application, that is, the ~/.origen will be created, bundler will populate all the gems and the lbin, etc. That is, my home Origen directory now looks like this:

image

Plugins can then add a 'global_commands' to the shared launcher in its config/application.rb file. This behaves exactly the same way as the shared_commands.rb. This only difference being global_commands will be available outside an application and shared_commands will still require an application. When run inside an application, global will be available as well though.

In pictures, this what I'm talking about:

(Running within an application, with TE_Utils installed)
image

image

(Running externally to an application, with TE_Utils still installed)
image

This feature is optional though. Enabling it requires setting a site config variable to allow this. By default, Origen will behave as it currently does. So, users that are not familiar with this functionality do not need to use it.

I did have some success with combining Gemfiles and programmatically adding dependencies. I opted to not do this because it becomes a distribution nightmare as an Origen application is now not dependent only on the Gemfile that's included. Additionally, we can't use Bundler for this, so we need to further build in utilities to handle getting a gem, checking its dependencies, etc., basically going back to the .imports that we abandoned once already in favor of something tried-and-tested. Otherwise, we would need to preprocess all Gemfiles, check the dependencies, etc. So, this note is really just saying: yes, we could combine multiple dependency sources, but I elected not to because it’s a lot of work to just end up violating our desire that a single application names all its dependencies.

This pull request is just to show my view of how Origen should start up to allow global dependencies and allow Origen to be used as a tool-dispersal platform in addition to its regular purpose. If this is accepted, I plan to take on adding the following as well:

	1) Document this new process on the Origen docs and include instructions for creating and editing a user site config, and the changes to LSF (see further down).
	2) Add a feature for origen to create a user site config upon start up (and have it be empty, or have commonly-used features commented out or set to false).
	3) Tie into either Origen 'new' or the app generator itself to allow for custom app generators.
	4) Add a way to add 'default' gems to the Gemfile upon start up. This would essentially allow a company's IT (or other toolset/origen expert) to setup a default site config and a default Gemfile, allowing for less experienced users to bypass this feature's setup altogether but still get the benefits. I plan this to be something along the lines of a site config path to a default Gemfile that will be compared at start up and copied and rerun if needed.

I say this because I didn't want to add all of these but then have this feature declined. Not sure how the team feels about Origen being multi-purposed into a tool distribution platform as well, or if this should be the way to go about it. This was the most natural way I could think of to take advantage of Bundler and Ruby's built-in dependency features while maintaining an 'Origen' feel and ultimately give full authority to if and how it's used to the individual users.

In order to do this I had to slightly alter how Origen.root behaves. Some plugins (like testers) uses the plugin manager in application.rb, which could now be loaded even when not run in an app. Using the plugin manager uses Origen.root, which would fail with a lookup error when run outside of an application. In this PR, if Origen.root would normally fail, Origen.top will be returned instead. This has the bonus of making the origen_updater plugin happy, which was having some issues as well when being run globally. Origen.root! Stayed the same for now as I think that one should always work, since an application can't call it (there's no application), a plugin would have it point to itself, and Origen core doesn't use it from what I can see.

Added method running_globally? (aliased as invoked_globally?) that is just the opposite of in_app_workspace?

Also added a global LSF and LSF manager. This is the same as the application's LSF/LSF manager, but available when running globally. The log directory for this will be Origen.root, which is now ~/.origen, so ~/.origen/.logs. Since Origen.lsf method was already taken though, and changing that would break the API and a lot of plugins, I added Origen.lsf! to get the global @lsf object. Origen.lsf will now pick between the application's lsf_manager or the global lsf_manager, depending on where Origen is invoked.

The TL;DR of this pull request is I want to use Origen also as a means of general tool (i.e. scripts) distribution, but in order to really be able to do anything you must be running from an Origen application, or else have your site manager configure Origen for you and cease using Bundler. This pull request aims to allow users to customize their own global installs and allows for plugins to add global commands that can be run outside of an application.


ginty:
Thanks very much @coreyeng, this is very timely.

I haven't read through the details here yet, but this is definitely something I think we need and in fact I wanted to add something very similar to allow the origen_sim plugin to be able to add a global command that verification teams could run to create an Origen testbench for a design without having to go through the process of setting up an application.

I'll read through it and give you more feedback soon. Thanks again!

coreyeng:
I fixed one of the travis builds but it looks like its now failing with the same error @info-rchitect had on the last commit. I pulled in the commit so makes sense that I have that same error too. I don't really understand the error though. Is this just travis/rvm not working properly? I don't see how something we checked in would cause this.

ginty:
Seems like a travis error, potentially you could try removing that ruby from .travis.yml, we should probably just drop it and replace it with the most recent Ruby (whatever that is)

info-rchitect:
Not sure I understand it all yet honestly.  @coreyeng could you give some examples?

coreyeng:
@info-rchitect The only example I really have is the one in the PR. Do you just want more details for what I changed in the start up? Or for the global commands? Or one of the other things I changed?

ginty:
Hi @coreyeng, could you elaborate more on how TE_Utils gets loaded?
I can't work out from the description  how you actually add a plugin like TE_Utils to the global installation and make it available to other users.

coreyeng:
Sure. When Origen is first sourced, bin/origen can now check if an lbin directory exists in the origen_home_dir location (~/.origen) by default. If it does it will use that. That is, ~/.origen/lbin/origen executable. I use Bundler to do that, so I also initialize a Gemfile. My home (or global) Origen install now somewhat resembles that of an application. That is, I have a lbin, .bundle, Gemfile, Gemfile.lock, and the new .session directories, in addition to the gems directory.

This Gemfile starts with just including Origen, nothing more. But, I can add additional gems to it, or point to a specific version of Origen or to a local path. For example, my current Gemfile (I was testing the startup of origen_testers as well) looks like this:

image

This file is in ~/.origen/Gemfile. Bundler will now use this. If I go to the ~/.origen/ and run Bundle I can see the new gems have been added.

image

In this way I can add whatever gems I want to my global install. Or point to specific Origen version or a local copy. The bin/origen file still is in command though, it is still the starting point. The proper site config variables need to set and a Gemfile present then the /bin/origen will switch over to running that Gemfile for non-application Origen invocations.

The global_commands.rb file then acts in much the same way as the commands.rb file does with a plugin's shared_commands.rb. My config/application.rb file in TE_Utils now has this:

image

And I added a config/global_commands.rb file that looks very similar to the config/shared_commands.rb file:

image

I hope that gives a better explanation of what I did to add global gems. If not, I can try to provide more details, or do some kind of demo. I think the big thing is just that the ~/.origen before only had 'gems' now has its own Gemfile, .bundle, and lbin that can be used when no application is present.

ginty:
Thanks, that really helped, I get it now and I like it.

A few comments on things that spring to mind:

- Do we really need the site_config-based enable? I understand that adding origen_install_dir as a new site config option makes sense, but why do we need the additional enable? Why not just load the user's global setup if present, what would be the need to turn it on and off?
- Could we rename origen_install_dir to something else, maybe just install_dir or global_install_dir? The reason is that all site config parameters can be overridden in a user environment by using the same param name in upcase prepended with ORIGEN_, so ORIGEN_ORIGEN_INSTALL_DIR is a bit ugly.

Can you think about how we can incorporate a true global installation as well?

I definitely want to make the origen_sim plugin globally available within NXP, so that we can make verification engineers able to run 'origen sim:some_command' outside of an application but also without having to setup their home environment like this.

I think we need to have 3-levels where plugins can be added:

~~~
global  # May be similar to what you have here, but would be done in central area like our TR
  |
  |___ user_global    # What you have here
          |
          |___ application
~~~

coreyeng:
For the first bullet, I had it be disabled by default as I didn't know what the opinion on this would be and it would be something I could set up for myself and our PEs but not affect anyone else. I also wanted to give the most flexiblity. We can take it out, change it to be enabled by default, change the name of the variable, whichever. Running if present wasn't some thing I was considering because I was actually having origen build the home directory if the enables were present but the home directory wasn't.

For the second bullet, sure. We can rename it to whatever. I only put about five seconds of thought into those names anyway. :p

For a true global install, depends on what exactly we want. I think the easiest would be to just create a new install somewhere in the tools directory and edit that gemfile. Verification can source that install instead of our normal one (and shut off the user install since it wouldnt be needed :D). This would allow you to maintain whatever verification is running, if that's what you want.

Alternatively, using the site config feature you added recently with the home install, they could basically just source a script in the tools directory that would set the global_install_dir to some prebuilt install, that you or IT would maintain.

More alternatively, they could source a setup that references a site config with a specific gemfile to use. They could then build their global origen in their home directory without needing to do anything themselves. Then they can further play with that if they want, or leave it alone and still be fine. I still need to add that feature though.

Even more alternatively, I can continue to think about it. For this PR, I was assuming a 'standard' user, that can't just go edit the global Gemfile manually or install origens wherever. I don't know if there are some conventions we should working within when making things to run within TR?

Just for clarity, you aren't talking about trying to combine dependencies, right? Just different ways of 'picking' which dependencies  (Gemfile) it should use? I experimented with that but couldn't get a good, error-proof way of combining them. Unless you know of a good way to go about it.

ginty:
> Running if present wasn't some thing I was considering because I was actually having origen build the home directory if the enables were present but the home directory wasn't.

OK that sounds good, I hadn't considered that.

> Just for clarity, you aren't talking about trying to combine dependencies, right? Just different ways of 'picking' which dependencies (Gemfile) it should use? I experimented with that but couldn't get a good, error-proof way of combining them. Unless you know of a good way to go about it.

Yes, I agree that trying to actually cascade dependencies will be difficult, so for clarity:

- In an app workspace, you always get what the application's Gemfile specifies
- Outside of a workspace, you get the user's own global setup if it is enabled
- If it is not enabled, then you get the true global setup if present, if not it works as today

So I guess that means we do need an enable switch on the user's global setup after all.

For the sake of clarity, let's call the setup in the home directory the "user setup" and the one in a central area accessible by all users as the "global setup".

The global setup would have exactly the same structure as the user setup, the only difference is that it would be managed by a single person, but it would be done in a central area that would be visible to all users. In NXP, that means it would live in a TR package.

By being centralized it means that we can push a default set of tools to all users, but it is likely to be slower to pick up new versions of plugins.
Power users, or those who need access to the latest version of a plugin right now, can enable their user setup and then take full control.

Does that make sense?

What about the following new site variables:

user_install_dir (what you currently call origen_install_dir), set to ~/.origen by default
global_install_dir, set to nil by default, in NXP we will set this to a path that points a TR package
user_install_enable, set to false by default, setting to true will enable the user setup instead of the global one

Can you add support for that to this PR?

Thanks!

coreyeng:
Okay. Yes, that all makes sense. I'll get started on adding those and update the PR when complete. I'll try to sneak in doc updates with this as well.

coreyeng:
Updates:

Added feedback from @ginty to PR with one exception: as I writing the docs, I found myself using the word <code>global</code> often and it was starting to lose meaning. Additionally, I ended up using <code>universal install</code>, to refer to the actual IT-managed installation (like what we have in NXP) so then universal vs. global install was getting convoluted. So, since the global install is really geared towards being used in a TR (tool-repository) package, I renamed <code>global</code> to <code>tool_repo</code>. I.e., <code>global_install_dir</code> to <code>tool_repo_install_dir</code>. I can change it back to <code>global</code> though, but I think it helped the docs flow better and not use <code>global</code> to mean few different things.  

Updated docs. I added a new page to the 'getting started' section about invocations, updated the plugin creation guide to discuss sharing global commands, and added an 'advanced topics' section to talk about the different installation options using the site config. I put this in a new section since I think it's out of the scope of just 'getting started' but also kind of an off-the-wall feature that may just clutter up the 'misc' section. I can move it though if needed. 

Small fix to the 'fix_my_workspace' script since it always wanted to use the universal install.  Added application options to global commands since I forgot about this last time.

I found that if a YAML file exists but is empty of any parameters, the load will 'succeed' but the results is nil. I.e., an empty site config add the value 'nil' into the site-configs arrray, causing Ruby to explode. Added fix for this.

Made some changes to the site config to give <code>gem_install_dir</code> and <code>user_install_dir</code> some dynamic default values without needing to evaluate anything from the site config.

Origen.home now points to <code>hom_dir</code> instead of directly to ~/.origen, even though most of the time it will evaluate to there.

bin/fix_my_workspace
```diff
@@ -4,8 +4,17 @@ $VERBOSE = nil  # Don't care about world writable dir warnings and the like
 if $_fix_my_workspace_version_check
   $_fix_my_workspace_version = '0.5.0'
 else
-  require 'origen'
-
+  # if we're running this from a development environment (i.e., bin/ inside of origen repo) then
```
ginty: This needs to be changed upstream in the Origen updater plugin, once an app updates to the newer version that plugin it will automatically mirror the changes to the local copy of fix_my_workspace like this one.

coreyeng: I updated my Origen to use the latest updater but it still didn't work. The error I'm getting is:

<code>
/pkg/fs-ruby-/2.3.0/x86_64-linux/lib/ruby/gems/2.3.0/gems/origen-0.11.0/lib/origen.rb:280:in `root': Something went wrong resolving the application root! (RuntimeError)
</code>

Does this need to be fixed back at that install of Origen? That's not the version I'm sourcing either. But the Origen.root issue is something that I've corrected in this PR. Doesn't seem to be being used by the fix_my_workspace unless I explicitly load it, as in the changes above.

It seems like, since that's a standalone script, its using the Origen version installed in the system gems, not the version I'm pointing it to.

ginty: Probably I wasn't clear, but the point is that if this change is introducing something that breaks the fix_my_workspace script, then we need a parallel update to the master version of that script to make it work with your update but which is also backwards compatible with existing Origen.

This is the master version here: https://github.com/Origen-SDK/origen_updater/blob/master/bin/fix_my_workspace

You should never make edits to this file within an application like this, since they will get overridden by updating the version of origen_updater.

templates/web/guides/advanced/invocations.md.erb
```diff
@@ -0,0 +1,129 @@
+% render "layouts/guides.html" do
```
ginty: This is a great guide!

origen_site_config.yml
```diff
@@ -2,11 +2,51 @@
...
+#  not existing (or no 'startup app' exists there), Origen will raise an error.
+
+# By default, don't allow user install. This is more for power users to utilize for debug or who want absolute control
+# over their environment.
+user_install_enable: false
+
+# Default to the user's install dir being the .origen in their home directory but allow it to be overridden
+#user_install_dir: ~/.origen
+
+# Default 'global_install_dir' to nil, meaning no 'global_install_dir' is present. If neither this nor user_install_enable
```
ginty: No issue with using the tooL_repo terminology, just to note that the comments in this file need updated to reflect the new names

ginty:
This is a really great update @coreyeng, thanks a lot.

Do you have a demo of it you could show me before we finally merge?

bin/fix_my_workspace
coreyeng: Okay. I understand. I didn't know there was another version in origen_updater. I was the running the one directly in <code>bin</code> directory of Origen. Why do we have two of these? Is the usage of the Origen core one deprecated and I just didn't know it.

I don't think this PR breaks the script per se, I think its just a bug in the script which makes debugging a bit wonky. I think that since this is a standalone script, requiring 'origen' is actually requiring the system gem, so the changes I made to Origen aren't being registered unless I explicitly require them. But then, only a few files were included (i.e. only the ones I need :p). I should take a better look at this and make it more robust. Whether this needs to be in this PR or a future improvement is up to you. I guess this will go in the origen_updater anyway so wouldn't be for this PR.

Ultimately, this would work itself out if our system gem origen updates to this PR's version. Then it's the same end result.

origen_site_config.yml
coreyeng: Oops. My bad. I'll correct the comments and update.

coreyeng:
Sure. I can do a demo. Let me know a time that works for you, but I'll be out from 8/3 to 8/8. Could probably still fit in something in the morning (Austin time) though. Or today if you're available.

ginty:
Cool, let's just do the demo next week when you get back, probably don't want to release something like this beforehand if you are going to be away anyway.

bin/fix_my_workspace
ginty: The fix_my_workspace script is owned/provided by the origen_updater plugin. The reason that it is not exposed as a regular shared command is that it has to be available at times when something has gone wrong with the user's workspace and they cannot invoke origen.

I went back and forth over a few different ways to guarantee that the user would always have access to the script, but in the end I felt the less magic the better and it was hard to beat just copying it into the application's bin directory.
That meant they can always just run ./bin/fix_my_workspace directly without having to go through Origen or Bundler's dispatch systems.

I still wanted to manage the script in a central location of course, so every time origen is invoked within an application that includes origen_updater, it will have a quick check to see if the version of the script they have locally is the one from the current version of origen_updater, if not it will copy it over on the fly.

Now ideally, I would prefer not to have required origen at all within that script. But since I needed access to the site_config to check for what system gems Origen should be using, I decided just to require it rather than try to re-implement the site config parsing within the script itself.
So far, that seems to have worked ok, but there are probably things you need to consider here with the changes to Origen's dispatch.
I don't really understand the changes well enough to make a recommendation, you will know best yourself, but just think about we need to maximise the chances of the script working even if an application's workspace is broken such that "origen" cannot be invoked through Bundler in the normal way.

coreyeng:
Opened https://github.com/Origen-SDK/origen_updater/pull/1 to review changes to fix_my_workspace.

PR #106, opened 2017-07-20
Added a String core extension to convert Verilog numbers, represented…

info-rchitect:
… as Strings, to straight bit strings and to Integers

### DIFF
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -55,6 +55,7 @@ unless defined? RGen::ORIGENTRANSITION
     autoload :Netlist,   'origen/netlist'
     autoload :Models,    'origen/models'
     autoload :Errata,    'origen/errata'
+    autoload :Fuses,     'origen/fuses'
 
     attr_reader :switch_user
 
--- a/lib/origen/core_ext/hash.rb
+++ b/lib/origen/core_ext/hash.rb
@@ -64,7 +64,8 @@ class Hash
           # ...push that on to the list of places to search.
           stack << v
         elsif key.is_a? Regexp
-          search_results[k] = v if key.match(k)
+          binding.pry if $stop == 1
+          search_results[k] = v if key.match(k.to_s)
         else
           return v if (k == key)
         end
--- a/lib/origen/core_ext/hash.rb
+++ b/lib/origen/core_ext/hash.rb
@@ -49,4 +49,23 @@ class Hash
   def longest_value
     values.map(&:to_s).max_by(&:length)
   end
+
+  def recursive_find_by_key(key)
+    # Create a stack of hashes to search through for the needle which
+    # is initially this hash
+    stack = [self]
+    # So long as there are more haystacks to search...
+    while (to_search = stack.pop)
+      # ...keep searching for this particular key...
+      to_search.each do |k, v|
+        # ...and return the corresponding value if it is found.
+        return v if (k == key)
+        # If this value can be recursively searched...
+        if v.respond_to?(:recursive_find_by_key)
+          # ...push that on to the list of places to search.
+          stack << v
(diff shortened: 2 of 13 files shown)


ginty:
Hi @info-rchitect, sorry it took a while to look at this, thanks as always for the PR.

I'm wondering if we really need to be introducing a new method here, verilog_to_i?
Having the user identify when the source number is/is not in verilog format doesn't feel very Rubyish.
Could we not enhance the existing to_dec method to identify the leading verilog identifiers?

~~~ruby
"0xF".to_dec      # => 15
"hF".to_dec       # => 15
"4'hF".to_dec     # => 15
"0b1000".to_dec   # => 8
"b1000".to_dec    # => 8
"4'b1000".to_dec  # => 8
~~~

info-rchitect:
i will take a look at integrating the 2 methods.  can you take a look at the build error.  When I run the fuses_spec test alone it passes and only fails when run with other spec tests.  I understand the error message but can't understand why it is occurring.

ginty:
Thanks.
Yeah this is one of those cases where you need to use unload target to clean things up between tests.
This is also a good place to use an [Anonymous Target](http://origen-sdk.org/origen/guides/runtime/programming/#Anonymous_Targets) as you can keep everything self contained within the spec file.

I would normally do this for a spec suite that involves a top-level model:

~~~ruby
# Create a fresh DUT instance before each test, you could use before :all here if
# you wanted to use the same instance for all tests, but normally it is easier to debug
# and run tests standalone if you treat them all as standalone units
before :each do
  Origen.app.unload_target!
  Origen.target.temporary = -> { SoC_With_Fuses.new }
  Origen.load_target
end

# Good to clean up after yourself
after :all do
  Origen.app.unload_target!
end

it "can create and interact with top level fuses" do
  # Because the object has been loaded through a target, you can use the dut handle
  dut.fuses.size.should == 2
  dut.fuses[:ff1].name.should == :ff1
  # ...
~~~

info-rchitect:
Can I get an official release with this included tomorrow?  thx

ginty:
Thanks for the updates.

This one seems problematic to me:
~~~ruby
"10100110".to_dec.should == 166
~~~

Don't think you can assume that just because a string number only contains 1s and 0s that it is binary, I don't think that can be legal Verilog to do that either.

info-rchitect:
Yeah I was surprised to see that that was valid Verilog but from [this link](http://web.engr.oregonstate.edu/~traylor/ece474/beamer_lectures/verilog_number_literals.pdf):

Numbers are represented as:<size>’<signed><radix>value (”<>” indicates optional part)

If you know of a better source that contradicts this I will change it.  Heck, we can decide to make a stricter version of what a Verilog number string should look like.  Let me know...

ginty:
Nice link, but per first slide of that it says the default radix is decimal, should the above should be interpreted as a decimal number since it doesn't have a radix identifier.

info-rchitect:
i am fine with whatever but this sentence is why I did what I did:

_if no prefix given, number is assumed to be 32 bits_

ginty:
32-bits by default in the absence of the size identifier does not mean it should be interpreted as binary int he absence of the radix identifier too.
So it is saying "100" means a 32-bit number with value 100, not with value 4.

info-rchitect:
OK will make the update.  thx

info-rchitect:
Can this get pulled into an official release?

ginty:
Can you add a quick doc update to the model section to mention the fuse API?

ginty:
Also, adding a note on the support for Verilog formatted numbers to decimal would be good here: http://origen-sdk.org/origen/guides/misc/coreext/

PR #115, opened 2017-08-15
App generators

ginty:
This PR adds a new site config parameter to allow a company to add their own application starter templates.

The Origen app generators plugin will in due course be enhanced to provide an option of creating a new plugin that should be used to contain company-specific app templates/generators.

This update provides the Origen core hooks necessary for the 'origen new' command to be able to pick up such generators, either from an internal gem server, a file location, or from a Git repo.

A user's local copy of the generators will be refreshed every 24 hours or anytime they supply a --fetch option to 'origen new'. This ensures that they will always be pointing to the latest and greatest templates.

The ability to specify a previous version of the generators has been removed. It may be re-introduced in future if there is demand for it, but how to specify a particular version gets more complicated now that multiple generator sources are supported.

Documentation on company specific generators will be provided along with the main update to the Origen app generators plugin.

This PR also includes a small update to stop Origen showing the stack dump if the process ends via exit. This will generally make the command line output cleaner and means that we have the following rule for developers:

~~~ruby
# When exiting like this, no debug information will be provided to the user unless you also provide
# it via puts statements
exit 1 

# This type though will continue to show the stack dump in addition to anything else you provide
# via puts
fail "Something is wrong!"
~~~

### DIFF
--- a/bin/origen
+++ b/bin/origen
@@ -61,7 +61,7 @@ if origen_root
 elsif Origen.site_config.gem_manage_bundler && (Origen.site_config.user_install_enable || Origen.site_config.tool_repo_install_dir)
   # Force everyone to have a consistent way of installing gems with bundler.
   # In this case, we aren't running from an Origen application, so build everything at Origen.home instead
-  # Have two options here: if user_install_enable is tru, use user_install_dir. Otherwise, use the tool_repo_install_dir
+  # Have two options here: if user_install_enable is true, use user_install_dir. Otherwise, use the tool_repo_install_dir
   Origen.site_config.user_install_enable ? origen_root = File.expand_path(Origen.site_config.user_install_dir) : origen_root = File.expand_path(Origen.site_config.tool_repo_install_dir)
   unless Dir.exists?(origen_root)
     load File.expand_path('../../lib/origen/utility/input_capture.rb', __FILE__)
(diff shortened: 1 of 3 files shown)


ginty:
Thanks for the review @coreyeng.

Yeah something like the auto refresh is definitely required. I expect that in due course the creation of new generators by the internal NXP community will be a regular occurrence and we definitely don't want to make it necessary for the central Origen installation to be updated to roll them out. It could probably have been handled by your global system, but I see that as more of a power user's tool and I want to keep the creation of new applications as easy as possible, especially for new users.

The ability to enable a custom generator plugin via a path (or a Git repo) is not actually intended to be any kind of debug aid. In NXP we will never use those features and we will use the gem server based distribution. However, I wanted to add Git support since it may help other companies to use this feature before they get a gem server up and running, and when I was doing that it was easy enough to add a path option as well.

In terms of debug and development of your generators, all app generator applications will have the ability to test run the new app creation process within the application itself, thereby allowing you to locally test out new generators or proposed changes to existing generators. They will also all have an easy to setup regression test which will verify that all generators can build and that the generated app can boot and perform basic functions.
I'm currently working on the updates to the origen_app_generators plugin to enable that, and I'm adding a generators application as a new application type available from Origen core.

Your idea on registering the generators would probably work fine too. Don't think we need to worry too much about it though as integrating a generators plugin into the new command is very much a niche situation and it doesn't matter if it is a little bit non-std.

Let me know if you have any other comments after reading the above, otherwise I will probably merge this later today.

Thanks!

coreyeng:
Thanks for the response! That all makes sense to me.

PR #117, opened 2017-08-18
Add regex as possible input to regs() method

rlaj:
Convenience of being able to grab all registers of a model that match a given regular expression.

Example: set configuration register for all ADCs (i.e. adc0_cfg, adc1_cfg, etc.) 
dut.regs('/adc\d_cfg/').each do |k,v|
  v.write!(0x000A_5000)
end


ginty:
Can we add the example to one of the doc pages?

ginty:
Included in 0.23.0

PR #119, opened 2017-08-23
Few fixes

coreyeng:
PR just fixing some things that I've noticed:

1. Origen.current_user will create a new instance every time, meaning that our instance variables are lost. For example:

~~~
u = Origen.current_user
u.instance_variable_set(:@test, 'test')

u.instance_variable_get(@:test) #=> 'test'
Origen.current_user.instance_variable_get(@:test) #=> nil
~~~

I think that it’s a bit confusing the Origen.current_user creates a new user every time. I changed it to store the user in @current_user instance variable on Origen and return that if it exists. I added a 'refresh' option that can passed in to force current_user to be reevaluated if an Origen program SSH's as another user or something to that effect.

2. origen web compile, even though starting a development server, requires setting the mode to debug first.

I think this is just a bug since there's a line that says Origen.set_development_mode if remote is not selected, but this is called after the target is already checked, resulting in the load_target! method failing with modified files in the workspace. Instead, I added loading the target with :force_debug set to true if remote is not used. If remote is specified though, the target will still be checked for production worthiness.

It's simple enough to just use 'origen mode debug' to set the mode to debug but this was also a simple fix and I think is the actually desired behavior.

3.  require colored at origen top level.

This is an inconsistency from global vs. application commands. When running in an application, application/release.rb is autoloaded and thus colored is brought in. So, if running a global command from an application (as in during development of the global command), you don't need to have require colored in the app. But, when running globally, you'll get a ruby stack trace since colored isn't loaded. Also, if another gem that requires colored is loaded first, you won't see this error even when running globally, adding to the confusion.

This just adds colored at the Origen module to make global commands more consistent. 'fileutils' is also a culprit, so added that as well.

4. Updated design sync interface to handle newer-styled headers.

DesignSync 2016 header is of the form: '3DEXPERIENCER2016x', which isn't currently screened out. Added a case to handle this.

### DIFF
--- a/bin/origen
+++ b/bin/origen
@@ -178,7 +178,9 @@ rescue Exception => e
   # case the application code is responsible for printing a helpful error message.
   # This will intercept all other exits, e.g. via 'fail "Something has done wrong"', and split the stack
   # dump to separate all in-application references from Origen core/plugin references.
-  unless e.is_a?(SystemExit)
+  if e.is_a?(SystemExit)
+    exit e.status
+  else
     puts
     if Origen.app_loaded?
       puts 'COMPLETE CALL STACK'
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -4,6 +4,10 @@ unless defined? RGen::ORIGENTRANSITION
   require 'English'
   require 'pathname'
   require 'pry'
+  # require these here to make required files consistent between global commands invoke globally and global commands
+  # invoked from an application workspace
+  require 'colored'
+  require 'fileutils'
   # Keep a note of the pwd at the time when Origen was first loaded, this is initially used
   # by the site_config lookup.
   $_origen_invocation_pwd ||= Pathname.pwd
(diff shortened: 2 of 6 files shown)


coreyeng:
Also, there's a couple of failures of the specs I'm not understanding:

image

We always have the 'gem_manage_bundler' set to true, which is default. I'm not sure why that's expecting false. Also, we 7 plugins, not 6, so not sure if this a new spec or just didn't get updated when it was added.

I don't remember these specs so I didn't want to just change them. Or is there something I'm missing that's wrong with my environment?

ginty:
Hi @coreyeng, nice updates.

I've pushed a few additional fixes that deal with the test failures you were seeing:

- I noticed that the build passed here even though some of the tests were failing. I traced this to a bug in a change I made recently to stop it dumping the stack trace upon 'exit 1', but which meant that it was inadvertently swallowing the '1' exit code. So that's fixed now.
- The plugins == 7 was caused by me recently adding origen_app_generators as a remote so that I could include its documentation in the main Origen docs. Importing it in this way should not have registered it as an active plugin, but there was some code in the remotes manager which would automatically require any remotes that were Origen applications. I think this is not intentional behavior and comes from the fact that the remotes manager was originally developed from the original Origen plugin system. I have removed the offending code but would like @rlaj to OK it in case there is any consequence of this I am overlooking.
- The (recently added) site_config test worked fine stand alone but I didn't notice it didn't work as part of the full test suite because of the exit bug, so I've improved the test to work better within the context of the whole test suite.

By the way, I notice that your commits do not associate with your Github account, if you add your B50956@freescale.com address to your list of addresses in your Github profile it will fix that.

coreyeng:
@ginty Thanks for adding the specs fixes and for pointing out my account wasn't syncing. I didn't even notice that.

ginty:
@rlaj, can you review the changes to the remote manager here?
Would be good to include this in the next tag along with your updates...

ginty:
Included in 0.23.0

PR #121, opened 2017-08-25
metric conversion core extension

arkohler:
This commit adds some handy number->string conversion functions such as
100000.as_Hz => '100.0kHz'
0.00001.as_V => '10.0uV'

There are some spec fails that appear to have nothing to do with the files I've edited.


ginty:
Hi @arkohler, thanks a lot for this update, glad you found us!

Checking in Gemfile.lock is not strictly required for something like this, but its a good practice to get into for an Origen application and generally it helps to make sure that all developers have a consistent gem bundle, so I think we want to keep it under revision control.
Could you roll back the change to .gitignore and add Gemfile.lock back in?
Editing your .gitignore to take Gemfile.lock back out, then checking it and Gemfile.lock back in should do it.

After that we can merge this.

FYI the spec fail you see is being addressed in #119 

Also if you feel so inclined, you could add some documentation for this on this page: http://origen-sdk.org/origen/guides/misc/coreext/
The source for that is here: https://github.com/Origen-SDK/origen/blob/master/templates/web/guides/misc/coreext.md.erb

Wouldn't insist on that though and I don't want to put you off too much on your first PR!

Thanks!

arkohler:
Thanks @ginty.  I have reverted the commits to Gemfile.lock and .gitignore.  I had originally done that because something in the origen execution was updating Gemfile.lock and without committing the file (or ignoring it), some of the origen specs were refusing to run and aborting the execution.

ginty:
Included in 0.23.0

PR #130, opened 2017-09-06
Pipeline from application command extensions to origen core option parser

rlaj:
Allow applications (or plugins) to specify an 'option name' for command option extensions which, when set, will register in the Origen Core option parser.

Example:

This example shows how origen_testers can extend the 'generate' command with a '--compile' option which, when set, will register in the Origen Core option parser as *testers_compile_pat: true*.  In turn, that can be passed along to various processes associated with the base command (e.g. can be passed to the listener pattern_generated for post-pattern pattern compilation).

origen_testers: config/shared_commands.rb
```ruby

case @command
  when "generate"
    @application_options << ['--compile', 'Compile the current pattern or the list', { option: 'testers_compile_pat' }]

```

### DIFF
--- a/lib/origen/commands/compile.rb
+++ b/lib/origen/commands/compile.rb
@@ -43,10 +43,13 @@ Usage: origen compile [space separated files, lists or directories] [options]
   opts.on('-r', '--reference DIR', String, 'Override the default reference directory') { |o| options[:reference] = o }
   opts.on('-z', '--zip', 'Gzip all output files (diff checking will be skipped)') { |o| options[:zip] = o }
   app_options.each do |app_option|
-    ao = app_option.include_hash_with_key?(:option)
-    if ao
-      app_option.delete(ao)
-      opts.on(*app_option) { options[ao[:option].to_sym] = true }
+    if app_option.last.is_a?(Proc)
+      ao_proc = app_option.pop
+      if ao_proc.arity == 1
+        opts.on(*app_option) { instance_exec(options, &ao_proc) }
+      else
+        opts.on(*app_option) { |arg| instance_exec(options, arg, &ao_proc) }
+      end
     else
       opts.on(*app_option) {}
     end
--- a/lib/origen/commands/compile.rb
+++ b/lib/origen/commands/compile.rb
@@ -1,6 +1,8 @@
 require 'optparse'
 require 'origen/commands/helpers'
 
+include CommandHelpers
+
 options = {}
 
 # App options are options that the application can supply to extend this command
--- a/lib/origen/commands/compile.rb
+++ b/lib/origen/commands/compile.rb
@@ -1,4 +1,5 @@
 require 'optparse'
+require 'origen/commands/helpers'
 
 options = {}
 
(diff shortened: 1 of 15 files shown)


ginty:
Hi @rlaj,

Thanks for the doc update, however upon reading it I thought this API might be hard for others to understand, and then I got to thinking that we should also try and add support for negatives and other more custom use cases.

I think changing that last argument to a block instead of a hash might make it easier to understand what it is doing, so equivalent to your initial example:

~~~ruby
@application_options << ['--compile', 'Compile the current pattern or the list', -> { options[:testers_compile_pat] = true }]
~~~

But it can now also support non-booleans and anything else:

~~~ruby
@application_options << ['--compiler_dir', 'Use the given compiler directory', -> (dir) { options[:compiler_dir] = dir }]
~~~

If no block is given as the final argument it would do the existing behavior.

Now, the implementation is easier said than done! But I think that is a good API from a user perspective.

The trick is that when the block is called, it needs to run with the context (the binding) of where the option parsing is being done, and not from the context where this additional option is being defined, i.e. where 'options' will refer to some other variable.

This is untested, but something like this should work:

~~~ruby
if app_option.last.is_a?(Proc)
  proc = app_option.pop
  opts.on(*app_option) { instance_eval(&proc) }
else
  # Original implementation
end
~~~

Using instance_eval to execute the block should hopefully mean that `options` ends up referring to the local `options` hash within the parser.

Additionally though, we should support capturing the variable and passing that to the block if required.
Probably sniffing the proc object as shown here will work: https://stackoverflow.com/a/10181491/220679

ginty:
Was thinking, this might be easier to implement and is probably just as good, up to you:

~~~ruby
@application_options << ['--compile', 'Compile the current pattern or the list', -> (options) { options[:testers_compile_pat] = true }]

@application_options << ['--compiler_dir', 'Use the given compiler directory', -> (options, dir) { options[:compiler_dir] = dir }]
~~~

rlaj:
hi @ginty, embarrassingly, that took me a lot longer to figure out than it should have.  please take another look

ginty:
@rlaj, don't worry about that, its easy to write code snippets into a comment box that don't actually work, thanks for taking the time to work out the details, this is a good update.

lib/origen/commands/compile.rb
```diff
@@ -43,7 +43,16 @@
   opts.on('-r', '--reference DIR', String, 'Override the default reference directory') { |o| options[:reference] = o }
   opts.on('-z', '--zip', 'Gzip all output files (diff checking will be skipped)') { |o| options[:zip] = o }
   app_options.each do |app_option|
-    opts.on(*app_option) {}
+    if app_option.last.is_a?(Proc)
+      ao_proc = app_option.pop
+      if ao_proc.arity == 1
+        opts.on(*app_option) { instance_exec(options, &ao_proc) }
```
ginty: Do we even need the instance_exec, given that we are passing in the options arg?

This would be a more common approach, though I don't know what instance_exec does, maybe its the same:

~~~ruby
opts.on(*app_option) { ao_proc.call(options) } 
~~~

rlaj:
@ginty, can you take a quick look at it again? In particular, the way I'm reusing the code.  Doesn't feel quite right to have the method just sitting there and not part of any object.

lib/origen/commands/compile.rb
rlaj: Yes, that works too.  I switched it over.

rlaj:
Well, the CI failed on it, so guess it needs more work...

ginty:
hi @rlaj, yeah the commands in general are a bit bad practice how its all just sitting there in the global namespace.

The main issue here is that after those helpers are loaded I think it means that every object will respond to extend_options, though not sure?

Two ways I would approach it to clean it up:

* If dumping stuff into the global namespace like this, I always try and make the method name start with an underscore, so make it _extend_options for example. Sometimes I also do that for methods that should be private but for one reason or another I can't actually put them behind the private keyword.
* Along similar lines you could wrap your helper methods with `module Origen module CommandHelpers`, then you can call `Origen::CommandHelpers.extend_options`, which is probably the best approach in this case.

ginty:
hi @rlaj, yeah it seems a bit messy right now with various approaches to including/extending it into the commands.
I think it should be this:

~~~ruby
# lib/origen/commands/helpers.rb
module Origen
  module CommandHelpers
    # Note the self, to make this a module method
    def self.extend_options(opts, app_opts, options)
    end
  end
end

# lib/origen/commands/generate.rb
require 'origen/commands/helpers'

# No need to include it, in fact you don't want to mix it into the global scope

Origen::CommandHelpers.extend_options(opts, app_options, options)
~~~

rlaj:
Updated.

PR #131, opened 2017-09-07
Update pattern generation common api with overlay/capture information

pderouen:
Web guide update to explain how to use tester implemented overlay/capture features added in origen_testers 0.10.0


ginty:
Looks great, thanks!

PR #133, opened 2017-09-11
New 'origen gem' command set

rlaj:
Simple set of commands to get current gems being used or to fetch a copy for quick debug.

```text
Usage: origen gem
       origen gem gem_name [option]
       origen gem fetch gem_name 
       origen gem clean (gem_name|all)

Quickstart Examples:
  origen gem                           # Displays the list of currently used gems
  origen gem gem_name                  # Displays details of specified gem
  origen gem fetch gem name            # Copies gem source to a local repo (<application_top_level_path>/tmp/gems)
  origen gem clean (gem_name|all)      # Removes/deletes the local copy of the gem source
```

The gem list includes name, version and source locatiom.
The fetch command will copy the source code to a local directory at <application-top>/tmp/gems.


ginty:
No auto update of the `path:` in the Gemfile on gem fetch/clean?

chrisnappi:
Wow - this was just a question to get the SO number up - thanks!

Is it possible to use the info in application.rb to pull via rc rather than copy?

ginty:
That's a good point, I was always thinking that we had to have a central place to look that up, but yes we could get it from within the application: `Origen.app(:origen_swd).config.rc_url`

Of course that would only work for Origen plugin gems, but that's probably 99% of the use case, and if you can't resolve the url that way we could fall back to this copy from local approach.

rlaj:
@ginty, right now, none of these commands actually re-direct the application to used the fetch version, it still requires a manual edit to the Gemfile after the copy is done.   I had envisioned doing the re-direct automatically, but the complexity level shot up pretty quickly from there, and once I started thinking about corner cases, I was a little hesitant to disturb the Bundler configs on the fly.  But I can re-visit if you think there are enough use case(s).

@chrisnappi, agree with above post, shouldn't be too difficult to add

ginty:
I think if we're going to the trouble of adding a command for this we should try to automate the whole procedure.

If I was going to do something like this, I would leave markers behind after doing edits so that they are easy to undo.
Example lifecycle:

~~~ruby
# Gemfile
gem "origen_swd"
gem "origen_testers"
~~~

`origen gem fetch origen_swd`

~~~ruby
# Gemfile
#COMMENT-origen_swdgem "origen_swd"
gem "origen_testers"
gem "origen_swd", path: "path/to/my/version" # ADDED-origen_swd
~~~

`origen gem clean origen_swd`

~~~ruby
# Gemfile
gem "origen_swd"
gem "origen_testers"
~~~

The rollback then becomes pretty simple to remove lines containing `"ADDED-#{gem_name}"` and deleting "#COMMENT-#{gem_name}".

rlaj:
Ok, I'll update.

But I wasn't even considering doing a file edit to the Gemfile.  I was headed towards doing something with session variables and pre-empting the Bundler load, which seemed a bit more dangerous.

chrisnappi:
Yeah, trying to mess with bundler would be dangerous - I think editing the Gemfile should be fine though.  Thanks for this - I think it will aid development and prevent hacking.

coreyeng:
Maybe I'm alone here, but I think editing the Gemfile is more dangerous. Since its just Ruby, to make a robust Gemfile editor you'd pretty much need a Ruby parser to handle all the odd ways the Gemfile could be formatted. You'd have to handle additional options, odd spacing, comments thrown in the middle of the 'gem' and 'path' option, etc.  I think just returning a string to the console that says 'gem located at: ...' would be enough.

For Bundler, we could load the Gemfile back in bin/origen but before executing it, change the paths of some of the gems. 'Fetch' can set a flag somewhere (env/session variable, temp file, etc.).

I guess an 'add/remove gem' and 'change version' commands could be added as well then we basically say no manually touching the gemfile. This is probably more a pain though.

Maybe I'm looking too much into it... but I'm working on something similar for the YAML site config, but then I have a big red blob in the middle of the docs that says 'if you start manually messing with the config file instead of using Origen and it doesn't work, its on you'

chrisnappi:
Fair enough - I am also quite happy with it just returning an easily copy/pasteable result for the Gemfile.  You are probably right that either the Gemfile needs to be considered machine-generated or human-generated.  Having a mix is asking for trouble.

ginty:
Yeah good point @coreyeng, I guess my Gemfile entries tend to all be single line, but even just accounting for the fact that the existing args could span multiple lines makes them a bit harder to comment out.

This is quite an interesting gem which looks like it would help to create the necessary Ruby re-writer to do it robustly, but it might be overkill for this case though: https://thomasleecopeland.com/2016/11/17/rewriting-ruby-with-synvert.html

rlaj:
@ginty, @chrisnappi, @coreyeng  Changed to use RC for Origen gems, for non-Origen, just plain copy from source to local ./tmp/gems.

After digging around in the bowels of the bundler and rubygems gems, think I came up with a fairly simple Gemfile updater that will work with a generic Gemfile (even if it has some complex Ruby).  Basically, shove the overidden gem specifications to the top of the Gemfile, and then sort of added a  monkey patch filter to the 'gem' method to block any additional specifications for those 'session' gems.  Tried a few different cases and seems to work.  Please have a look.

```text
# ORIGEN GEM AUTO-GENERATED---------------DO NOT REMOVE THIS LINE-------------
# -- DO NOT CHECK IN WITH THIS SECTION!
# -- DO NOT HAND MODIFY!
# -- USE 'origen gem clean all' to reset
#
gem 'origen_jtag', path: '/path/to/application/my_application/tmp/gems/origen_jtag-0.15.0'
puts "\e[1;93;40mWARNING: Using session gem for 'origen_jtag'\e[0m"

def gem(*args)
  return if ['origen_jtag'].include? args[0]
  super(*args)
end
#
# /ORIGEN GEM AUTO-GENERATED---------------DO NOT REMOVE THIS LINE------------
source 'http://rubygems.org'
source 'http://rgen-hub.am.freescale.net:9292'

gem 'origen_doc_helpers', '>= 0.4.4'
gem 'byebug', '~>3.5'

gem 'origen_arm_debug', '1.0.0'
gem 'origen_jtag', '0.15.0'

gemspec
```

ginty:
Hey @rlaj, I've had a play around with this - it's very slick and a great solution you came up with!

I notice that the gem command needs to be added to the blurb that is displayed when you just run `origen`.

Otherwise my only slight reservation is that I'm not sure if we should be introducing a new top-level command given that we already have the plugin command. It seems it might be confusing to have them both going forward.

Do you or anyone else have any thoughts on how we might combine them into one?

coreyeng:
@ginty I don't really think of a 'plugin' and a 'dependency' as the same. I think of a dependency as just a passive asset that needs to be included, but a plugin as a dependency that passively includes stuff can be actively change the functionality of Origen (changing the pattern header, pattern dispatcher, etc). I'm think having two separate commands is fine. I also think that the 'origen pl added' should be a bit more selective in what it returns. I haven't used it in a while but back in RGen days it just returned all of our internal gems, regardless of what they did. How exactly the 'plugin added' should define a plugin... I'm not sure. But that's my vote.

@rlaj The only concern I have is that I think this will allow for production releases to be done with 'local' gems. Since the Gemfile itself isn't changing, I think the local dependencies will slip through the cracks. Though, I'd actually advocate for changing the 'validate_production_ready' to use Bundler to validate that the release isn't local. I think you would just compare the path Bundler gives to wherever the bundle install directory is. Right now it just checks for 'path' in the line, so it wouldn't be very hard to fool as it is. https://github.com/Origen-SDK/origen/blob/0731f00a12a7737331fb85f6659706d9eb398495/lib/origen/application/plugins.rb#L21

ginty:
@coreyeng, I don't really recognize that distinction to be honest. To me, a gem means any gem, and then a 'plugin' is the subset of gems which are Origen plugins.
Given the way this new command works, that you can fetch any gem with it, the naming does make sense.

On the other hand, I'm sure that 99% of the time this fetch feature is used it will be used to fetch an Origen plugin. I think many might therefore expect `origen pl fetch jtag` to work, I know I would.

To me, putting this under the existing `pl` command therefore makes more sense. Sure, we are maybe stretching the truth a bit if it is used to pull in a non-Origen gem, but I'm not convinced that that minority use case justifies increasing the command surface area.

coreyeng:
I could see it in <code>origen pl fetch</code> as well. It still fits. Yeah, I think you're right, <code>origen pl fetch</code> probably makes the most sense and keeps the command listing more succinct, but I wouldn't mind though if it was under <code>origen gem</code> instead.

rlaj:
Working to finish up the pattern compilers in origen_testers, then will circle back and clean this up.  Think we can make it work within the pl command set.

rlaj:
Not coded yet, but here's the proposal.  Combine the proposed gem command set into the pl command set.  Update the 'pl added' to also display location and add option (--all) to display external gem dependencies in addition to Origen plugins.  Fetch and Clean work as they did in the 'gem' command set.  I will also add something to take care of @coreyeng concern above regarding production releases with local auto-modifications to the Gemfile.

Not sure where that leaves the 'origen fetch' commands, but I think the fetching really belongs here; tied to the plugins.

Please take a look at the proposed command set below:

```text
Usage:  origen pl
        origen pl [plugin name]
        origen pl [CMD] [options]

Quickstart Examples:
  origen pl                             # Displays the current plugin
  origen pl [plugin_name]               # Sets the specified plugin as current plugin
  origen pl reset                       # Resets the current plugin to none
  origen pl added                       # Lists the included Origen plugins
  origen pl added --all                 # Lists the included Origen plugins and external gems
  origen pl fetch [plugin_name]         # Copies plugin/gem source to a local repo (<application_top_level_path>/tmp/gems)
  origen pl clean [plugin_name|all]     # Removes/deletes the local copy of the plugin/gem source

The following commands are available:

  added                               Displays all plugins that are currently included in this app locally
  reset                               Sets the specified plugin as current plugin
  fetch                               Copies plugin/gem source to a local repo (<application_top_level_path>/tmp/gems)
  clean                               Removes/deletes the local copy of the plugin/gem source

The following options are available:
    -a, --all                        List all plugins/gems (including external)
    -d, --debugger                   Enable the debugger
    -h, --help                       Show this message
```

PR #136, opened 2017-09-12
Added in Origen::Tests module and made add_mode more flexible

info-rchitect:
```ruby
  add_test :your_test_id_1 do |t|
      t.description = 'your test description'
      t.conditions  = [:xminvdd, :maxvdd, :bin1_1500Mhz, :bin2_1300Mhz]
      t.platforms   = [:v93k]
      t.meta1       = 'dkwew'
      t.meta2       = 'jkjejkf'
    end
```


lib/origen/tests.rb
```diff
@@ -0,0 +1,31 @@
...
+        end
+      else
+        @_tests.recursive_find_by_key(expr)
+      end
+    end
+
+    def add_test(id, options = {}, &block)
+      @_tests ||= {}
+      if @_tests.include?(id)
+        Origen.log.error("PPEKit: Cannot create test '#{id}', it already exists!")
```
ginty: Looks like some app-specific reference to be removed

ginty:
I think this is a really good first step and generally I think moving test definitions into the model layer will yield a number of advantages longer term.

I assume that your vision would be for deeper integration of this in future? i.e. be able to feed the list of tests into the pattern generator?

We also need to think about how we reconcile/integrate this with the test program generator - i.e. your list of tests in the model here looks a lot like the list of tests in a test flow.
Are they the same or different, how do we converge them?

info-rchitect:
@ Ginty.  Yes you are correct that there needs to be strategic alignment between the test definitions and the tests mentioned in the flow.  Our current implementation is to import a test list with the following constraints:

1) Each row in the test list equates to a unique test pattern
2) Each row has a test conditions list (e.g. 1200Mhz, 5p2Gbps, etc.) detailing where the test needs to be executed in the flow.  We leave voltage conditions out of this list because those are typically set after characterization.

With those constraints I think the current definition will work well to feed into the pattern generator and flow generation.  I think the areas that need to be better fleshed out are how chip modes interact with tests.  Should tests be defined within the scope of a mode definition or just exists as a flat list like registers.

lib/origen/tests.rb
info-rchitect: fixed

PR #139, opened 2017-10-05
Added in Origen::PowerSupplies model

info-rchitect:
This should prove useful if and when an ATE levels API gets added to origen_testers.  When pins are added to the DUT, each can be assigned a :supply attribute.  I expect the PowerSupplies module will grow over time ti support things such as:

~~~ruby
dut.supplies(:vdd).pins # => [:pinA, :pinB]
dut.supplies(:vdd).has_pin? :pinA # => true
~~~

The PowerSupply initialization automatically creates 3 spec limits that are :min, :nom, :max (all are empty).  I am not really sure if I want to use a spec limit to store the actual power supply limits or not so feedback is welcome (they could just be straight accessors instead).

### DIFF
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -63,6 +63,7 @@ unless defined? RGen::ORIGENTRANSITION
     autoload :LSFManager,    'origen/application/lsf_manager'
     autoload :Fuses,     'origen/fuses'
     autoload :Tests,     'origen/tests'
+    autoload :PowerSupplies, 'origen/power_supplies'
 
     attr_reader :switch_user
 
--- a/lib/origen/model.rb
+++ b/lib/origen/model.rb
@@ -27,6 +27,7 @@ module Origen
       include Origen::Errata
       include Origen::Fuses
       include Origen::Tests
+      include Origen::PowerSupplies
     end
 
     module ClassMethods
--- a/lib/origen/power_supplies.rb
+++ b/lib/origen/power_supplies.rb
@@ -1,0 +1,33 @@
+require_relative './power_supplies/power_supply'
+module Origen
+  module PowerSupplies
+    def power_supplies(expr = nil)
+      if expr.nil?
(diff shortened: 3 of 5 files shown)


welguisz:
@info-rchitect 
I think that this will make 3 different ways to discuss power supplies within Origen.  Right now I know that the model includes:
*  lib/origen/pins/power_pins.rb
*  lib/origen/specs/power_supplies.rb

Is there a way for these changes to fit into one of them?  Or is there something completely different here?

info-rchitect:
Hi @welguisz ,

The power pins aren't a power supply, they are a in that would be connected to a power supply.  The second is something that looks custom tailored to the app you use.  I say that because, from a modeling perspective, a spec would not own a power supply.  By checking the Travis CI build fail report it looks like we have some namespace collision.  I think I will change mine to PowerDomain because that is really a more accurate name, as a power supply is ATE specific.  my intent is to model from the Design perspective.  Make sense?

welguisz:
@info-rchitect I just put power_supply.rb in specs as a temporary location until a better one was found.  I really don't want to have two diffrent domains for power supplies, one for documentation and one for test engineering.

I would say we could do the following:

1. Put a deprecation warning on Origen::Specs::Power_Supply
2. Add functionality that is in Origen::Specs::Power_Supply into Origen::PowerSupplies::PowerSupply

info-rchitect:
@welguisz that works for me.  Let me update the PR to the latest usage model.  could you please review the specs and suggest functionality edits?

welguisz:
@info-rchitect ... if you make the following change to Origen::Specs::Power_Supply and commit it, I can do edits inline

def intiialize (...)
  Origen.log.deprecated('abc')
end

info-rchitect:
@welguisz can you remove your commit so I can push a new clean commit?

info-rchitect:
i can't get this thing to work with the commits, will redo

PR #140, opened 2017-10-06
Added in Origen::PowerDomains module

info-rchitect:
~~~ruby
  def initialize
    add_power_domain :vdd do |domain|
      domain.description = 'CPU'
      domain.nominal_voltage = 1.0.V
      domain.voltage_range = 0.7.V..1.1.V
    end
    add_power_domain :vdda do |domain|
      domain.description = 'PLL'
      domain.nominal_voltage = 1.2.V
      domain.voltage_range = 1.08.V..1.32.V
    end
    add_power_domain :vccsoc do |domain|
      domain.description = 'SoC'
      domain.nominal_voltage = 1.5.V
      domain.voltage_range = 1.35.V..1.65.V
    end
  end
~~~

I think this module will help Origen develop an ATE levels API, help the folks using Origen::Specs for docs work, and will help ChipMode modeling in general by allowing users to dynamically change the power domain setpoint based on ChipMode.

~~~ruby
  it "can create and interact with power domains" do
    dut.power_domains.should == [:vdd, :vdda, :vccsoc]
    dut.power_domains.size.should == 3
    dut.power_domains.size.should == 3
    dut.power_domains(:vdd).description.should == 'CPU'
    dut.power_domains(:vdd).nom.should == 1.0.V
    dut.power_domains(:vdd).range.should == (0.7.V..1.1.V)
    dut.power_domains(:vdd).setpoint = 1.0.V
    dut.power_domains(:vdd).setpoint_ok?.should == true
    dut.power_domains(:vdd).setpoint = 0.65.V
    dut.power_domains(:vdd).setpoint_ok?.should == false
    dut.power_domains(/^vdd/).class.should == Hash
    dut.power_domains(/^vdd/).keys.should == [:vdd,:vdda]
  end
~~~

The second commit added this functionality;

~~~ruby

class SoC_With_Domains
  include Origen::TopLevel
  
  VDD_SIGNAL_PINS = [:pin1, :pin2, :pin3]
  VDD_POWER_PINS = [:vdd1, :vdd2, :vdd3]
  VDD_GND_PINS = [:vss1, :vss2, :vss3]

  def initialize
    VDD_SIGNAL_PINS.each do |p|
      $dut.add_pin p do |pin|
        pin.supply = :vdd
      end
    end
    VDD_POWER_PINS.each do |p|
      $dut.add_power_pin p do |pin|
        pin.supply = :vdd
      end
    end
    VDD_GND_PINS.each do |p|
      $dut.add_ground_pin p do |pin|
        pin.supply = :vdd
      end
    end
end

  it "can find pins that reference itself" do
    dut.power_domains(:vdd).signal_pins.should == [:pin1, :pin2, :pin3]
    dut.power_domains(:vdd).power_pins.should == [:vdd1, :vdd2, :vdd3]
    dut.power_domains(:vdd).ground_pins.should == [:vss1, :vss2, :vss3]
    dut.power_domains(:vdd).has_pin?(:vdd1).should == true
    dut.power_domains(:vdd).pin_type(:vdd1).should == :power
    dut.power_domains(:vdd).has_power_pin?(:vdd1).should == true
    dut.power_domains(:vdd).has_pin?(:pin1).should == true
    dut.power_domains(:vdd).pin_type(:pin1).should == :signal
    dut.power_domains(:vdd).has_signal_pin?(:pin1).should == true
    dut.power_domains(:vdd).has_pin?(:vss1).should == true
    dut.power_domains(:vdd).pin_type(:vss1).should == :ground
    dut.power_domains(:vdd).has_ground_pin?(:vss1).should == true
    dut.power_domains(:vdda).signal_pins.should == []
    dut.power_domains(:vdda).power_pins.should == []
    dut.power_domains(:vdda).ground_pins.should == []
    dut.power_domains(:vdda).has_pin?(:vdd1).should == false
  end
~~~

3rd commit added in some console sugar:

image


info-rchitect:
Looks like it is failing because I changed the gemspec fle.  how do I fix this?

ginty:
You need to run origen/bundle in your workspace, it will then update the Gemfile.lock, then check that in

info-rchitect:
@welguisz can you go ahead and add in your functionality?

info-rchitect:
so are we ready to go?  Any thoughts from the peanut gallery?

ginty:
Really nice PR @info-rchitect, thanks!

So is the plan for this to replace `Origen:Specs::Power_Supply`? If so, should we mark that as deprecated?

I really love the console output, we'll need to brush up the register tables to match.

One small suggestion to improve the console experience even further. You can define a custom class to replace the hash currently used by @_power_domains, this will then allow you to define a custom inspect method:

~~~ruby
class PowerDomainCollection < Hash
  def inspect
    # Your awesome table code here  
  end
end
~~~

Because it is a child class of Hash, you just use it like a hash in your code.

However, because it has the custom inspect method, you can just write `dut.power_domains` at the console and it will display the table.

Here is the inspect method which displays the regs in the console in a similar way: https://github.com/Origen-SDK/origen/blob/master/lib/origen/registers/reg.rb#L129

info-rchitect:
@ginty thanks for the compliments and the tips.  I will do another commit with the updated console code.  BTW, I think a future commit is to make all complex objects use the same inspect code such that we don't have varying output.  Couldn't we just override Object?

And yes this is meant to deprecate Origen::Specs::PowerSupply, do you have any examples of the Origen way?

Also, could you specifically address the concept that pins should always be owned by the $dut versus the individual power domains?  Is that a hard fast rule or just a guide?  Obviously, if people added pins to the power domains directly some code would need to be changed.

thx

info-rchitect:
one more thing regarding the inspect method.  I thought that just typing $dut.power_domains should return an array of power domain IDs, not the inspect table.  Can you mock-up how the dut.power_domains method behaviors a bit more?

info-rchitect:
Also, it was @arkohler who wrote the new fancy output code, on day 2 of learning Ruby.

ginty:
Here's an example of how to fire a deprecation warning: https://github.com/Origen-SDK/origen/blob/56bccd88078fcde4ad0f1f1e3b9dbac359d7be17/lib/origen/core_ext/string.rb#L64

Add something like that to the initialize method of `Origen::Specs::PowerSupply`.

I did wonder if `dut.power_domains` should just print an array of names to the screen, like `dut.regs` does.
Then I thought that the number of power domains is likely to be quite small, so why not just show table?
 Up to you though.

How to display something in the console is pretty unique to the class in question, can't really see how an Object-level inspect method would work.

I think you can add pins to any object today if you want. Adding them to the top-level is just a recommendation about how to manage them.

info-rchitect:
@ginty OK so I have been playing around with the inspect method and I see only one issue.  I cannot do the following:

~~~ruby
$dut.power_domains.each do |pd|
  puts pd.class
end
~~~

You must have gotten around this somehow right?

ginty:
Inspect has no effect on what the `power_domains` method actually returns.

Is the issue here that your code above is for dealing with an array, but you actually have a hash?
i.e. should it be...

~~~ruby
dut.power_domains.each do |name, pd|
  puts pd.class
end
~~~

info-rchitect:
@ginty let me do the commit, could you take a look?

ginty:
Remove the call to inspect and just return the PowerDomainsCollection instance in the same way as you did with the hash.

The console will automatically call inspect on what you return when you are operating from a console.

lib/origen/power_domains.rb
```diff
@@ -0,0 +1,29 @@
+require 'colorize'
+require_relative './power_domains/power_domain'
+require_relative './power_domains/power_domains_collection'
+module Origen
+  module PowerDomains
+    def power_domains(expr = nil)
+      if expr.nil?
+        if @_power_domains.nil?
```
ginty: Don't think you should return the keys here, this whole top half of the if branch could just be written as:

~~~ruby
@_power_domains ||= PowerDomainsCollection.new
~~~

PR #142, opened 2017-10-11
Fixed power_domain inspect output for some cases

arkohler:
The original printer portion of the inspect function assumes that all of the power domains have the same list of attributes to print out.  If any one of the attributes are missing, you could get glitches in the output.  This PR corrects that.  I've also added back in the voltage range attribute which was incorrectly excluded.

Here's an example of the original output were two of the domains are missing the setpoint attribute.
image

Here's the corrected output
image

### DIFF
--- a/lib/origen/power_domains/power_domains_collection.rb
+++ b/lib/origen/power_domains/power_domains_collection.rb
@@ -9,11 +9,11 @@ module Origen
         output_power_domain_list = []
         column_widths = {}.tap do |colhash|
           each do |domain_name, domain|
-            output_attr_list = []
+            output_attr_list = {}
             domain.instance_variables.each do |attr|
               attr_getter = attr.to_s[/\@(\S+)/, 1].to_sym
               attr_val = domain.send attr_getter
-              next unless [String, Numeric, Float, Integer, Symbol].include? attr_val.class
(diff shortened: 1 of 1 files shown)


ginty:
Looks good, want to add some test coverage for it?

If you add it to this file: https://github.com/Origen-SDK/origen/blob/master/templates/test/inspections.txt.erb

It will get compared to this output file during `origen examples`: https://github.com/Origen-SDK/origen/blob/master/approved/inspections.txt

Also, it doesn't have to be part of this PR, but it would be great if you could deploy your superior ASCII art skills to making the register tables equally pretty!

info-rchitect:
@ginty how do you add coverage?  Looking at the file it isn't completely intuitive to me.

ginty:
`origen examples` will compile the first file I linked to and compare it to the approved output (the 2nd file).

So, if you call inspect on a power_domains example within that file, the output will be compared to what you accept as the approved output going forward, therefore ensuring that it always works.

You can add a method to add power_domains to the $nvm instance used in the existing test case here: https://github.com/Origen-SDK/origen/blob/master/lib/c99/nvm.rb

info-rchitect:
we will do the examples very soon but think this can be merged.  the register console sugar will occur soon too.

PR #143, opened 2017-10-12
Fix regression manager's use of bundle exec

ginty:
I couldn't get the regression manager to work in one of my apps, I suspect probably related to some relatively recent changes to the Origen boot process.

Anyway, these changes fixed it.

It's basically removing the regression manger's use of `bundle exec` when launching origen commands whenever Origen is in charge of bundler (should be the case in 99-100% of apps).
Not sure why it used `bundle exec` originally, but the way it is now aligns to how you would invoke origen directly from the command line.


ginty:
Going ahead with the merge on this after no review comments, it is a low risk change and I believe is fixing a bug.

PR #147, opened 2017-10-18
Changed PowerDomain.voltage_range to unit_voltage_range because it…

info-rchitect:
… really refers to the range a unit could be fused to a specific value based on power/performance.  The PowerDomain class still needs to integrate spec limits that would handle the full range.  If anyone else has a better name for this attribute pls join the conversation.

### DIFF
--- a/lib/origen/core_ext/string.rb
+++ b/lib/origen/core_ext/string.rb
@@ -55,8 +55,17 @@ class String
   # Sanitizes the string for conversion to a symbol and returns a lower
   # cased symbol version of the string
   def symbolize
+    orig_had_leading_underscore = match(/^\_/) ? true : false
+    orig_had_trailing_underscore = match(/\_$/) ? true : false
+    new_str = gsub(/(\?|\!|\-|\/|\\|\n|\s|\(|\)|\.|\[|\]|-|{|})/, '_').downcase
+    # Get rid of adjacent underscores
+    new_str.match(/\_\_/) ? new_str = new_str.squeeze('_') : new_str
+    new_str.chomp!('_') unless orig_had_trailing_underscore
+    unless orig_had_leading_underscore
+      new_str = new_str[1..-1] if new_str.match(/^\_/)
+    end
     @@symbolize ||= {}
-    @@symbolize[self] ||= gsub(/(\?|\!|\-|\/|\\|\n|\s|\(|\)|\.|\[|\]|-|{|})/, '_').downcase.to_sym
+    @@symbolize[self] ||= new_str.to_sym
   end
 
   # acronyms
--- a/lib/origen/power_domains/power_domain.rb
+++ b/lib/origen/power_domains/power_domain.rb
@@ -1,7 +1,7 @@
 module Origen
   module PowerDomains
     class PowerDomain
-      attr_accessor :id, :description, :nominal_voltage_range, :nominal_voltage, :setpoint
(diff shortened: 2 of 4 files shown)


welguisz:
@info-rchitect how many changes are you going to make to PowerDomain ... I am trying to move PinTool import from using PowerSupply to PowerDomain.  Each time you make a change that hits upon attrs_ok?, I have to change code on my side.

IMO, if you have attrs_ok? you should have an options[:ignore_attrs_ok] to disable attrs_ok? so this does not become an issue.

info-rchitect:
@welguisz are you taking every release from origen automatically?  if so, I would advise against that, especially given the fact that there are [open pull requests](https://github.com/Origen-SDK/origen/pull/147) for this module.  Of course, I hope we close this soon but disabling the attrs_ok? would break other things.  Can you say what specifically you had to change?

info-rchitect:
All, after thinking about what nominal_voltage range/voltage_range is, it is really a unit_voltage_range.  power domains that have a :fixed unit_voltage_range apply the same voltage to all units.  power domains with a variable unit_voltage_range can program a range of voltages to the unit. What say you all?

welguisz:
@info-rchitect ... I am not taking every release from origen automatically.  I have some spare cycles and want to get this out of the way so I get less DEPRECATED warnings from PowerSupply.

What I was thinking with ignore_attrs_ok is the following:

In my file
~~~ruby
$dut.add_power_domain(:vdd, ignore_attrs_ok: true) do |domain|
  ...
end
~~~

in power_domain.rb
~~~ruby
def initialize(id, options={}, &block)
   ...
   fail unless attrs_ok?(options)
end

def attrs_ok?(options = {})
  return true if options[:ignore_attrs_ok]
  # regular code here
end
~~~

So my code runs without having to do checks.  I feel like attrs_ok? is locking down the experience to one implementation and narrows the scope.

info-rchitect:
@welguisz checking the attributes such that they are the correct type can't be disabled as it would essentially make the class be whatever the user wanted.  All of the other methods rely on attributes being of a certain type or value.  Was it the change from voltage_range to nominal_voltage_range (and now possibly unit_voltage_range) that caused an issue?  Given that I think you really don't care about that attribute, i can make it default to :fixed unless specified such that you should never have to program it on your side.  Pls tell me if that is not what caused your issue.

welguisz:
When the user sets options[:ignore_attrs_ok], the user should know that the class will not work as advertised.   My thoughts are "When you try to take out all risks, you add in lots of marginal, expensive costs with low returns."  Some one wiser than me said "the more idiot-proof you make something, the more idiotic your users become."

What about this as a compromise:

~~~ruby
def attrs_ok?(options = {})
   if (options[:ignore_attrs_ok])
      Origen.log.warning "Ignore attrs_ok for PowerDomain initialization.  By not having this, PowerDomains will not work as advertise."
      return true
   end
end
~~~

When PowerDomain hardens:
~~~ruby
def attrs_ok?(options = {})
   if (options[:ignore_attrs_ok])
      Origen.log.deprecate ":ignore_attrs_ok for PowerDomain initialization will be removed on MM/DD/YYYY. Please change your code before this day."
      Origen.log.info "Currently, attrs_ok? will return #{attrs_ok?}"
      return true
   end
end
~~~

info-rchitect:
@welguisz last commit should fix any issues you have

welguisz:
Thanks   ...

PR #157, opened 2017-10-26
Add block write!/read! ability to allow for better coding style for bit manipulations.

chrisnappi:
This pull request intends to add a method to provide for better coding style when dealing with large numbers of bit manipulations within a register.  These manipulations can be hard to read, and error prone when '!' is in play.

Previous code style:
```
  $dut.mtr.mct.reg(:bstart).bits(:bstart).write(0x4) # Program BIST but don't start them
  $dut.mtr.mct.reg(:bstart).bits(:cens).write(1) 
  $dut.mtr.mct.reg(:bstart).bits(:cenac).write(1)
  $dut.mtr.mct.reg(:bstart).bits(:brst).write(1)
  $dut.mtr.mct.reg(:bstart).bits(:cof).write(1)
  $dut.mtr.mct.reg(:bstart).bits(:clken).write(1)
  $dut.mtr.mct.reg(:bstart).write!
```
New coding style:
```
  $dut.wreg($dut.mtr.mct.reg(:bstart)) do |r|
    r.bits(:cens).write(1) 
    r.bits(:cenac).write(1)
    r.bits(:brst).write(1)
    r.bits(:cof).write(1)
    r.bits(:clken).write(1)
  end
  $dut.mtr.mct.reg(:bstart).write!
```

### DIFF
--- a/lib/origen/registers/bit_collection.rb
+++ b/lib/origen/registers/bit_collection.rb
@@ -523,7 +523,7 @@ module Origen
       def read!(value = nil, options = {})
         yield @reg if block_given?
         value, options = nil, value if value.is_a?(Hash)
-        read(value, options) if !block_given?
+        read(value, options) unless block_given?
         # launch a read reg
         @reg.request(:read_register, options)
         self
--- a/lib/origen/registers/bit_collection.rb
+++ b/lib/origen/registers/bit_collection.rb
@@ -502,14 +502,16 @@ module Origen
         result.to_s
       end
 
-      # Write the bit value on silicon<br>
+      # Write the bit value on silicon.
       # This method will update the data value of the bits and then call $top.write_register
-      # passing the owngin register as the first argument.<br>
+      # passing the owning register as the first argument.
       # This method is expected to handle writing the current state of the register to silicon.
       def write!(value = nil, options = {})
         value, options = nil, value if value.is_a?(Hash)
         write(value, options) if value
-        yield @reg if block_given?
+        if block_given?
+          yield size == @reg.size ? @reg : self
+        end
         @reg.request(:write_register, options)
         self
       end
@@ -521,10 +523,11 @@ module Origen
       #   reg(:data).read!         # Read register :data, expecting whatever value it currently holds
       #   reg(:data).read!(0x5555) # Read register :data, expecting 0x5555
       def read!(value = nil, options = {})
-        yield @reg if block_given?
         value, options = nil, value if value.is_a?(Hash)
         read(value, options) unless block_given?
-        # launch a read reg
+        if block_given?
+          yield size == @reg.size ? @reg : self
+        end
         @reg.request(:read_register, options)
         self
       end
--- a/lib/origen/registers/bit_collection.rb
+++ b/lib/origen/registers/bit_collection.rb
@@ -509,6 +509,7 @@ module Origen
       def write!(value = nil, options = {})
         value, options = nil, value if value.is_a?(Hash)
         write(value, options) if value
+        yield @reg if block_given?
         @reg.request(:write_register, options)
         self
       end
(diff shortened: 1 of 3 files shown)


ginty:
Like the idea, would this not be even better?

~~~ruby
$dut.mtr.mct.reg(:bstart).write! do |r|
    r.bits(:cens).write(1) 
    r.bits(:cenac).write(1)
    r.bits(:brst).write(1)
    r.bits(:cof).write(1)
    r.bits(:clken).write(1)
end
~~~

chrisnappi:
Yes, that would be better.  Not sure how to implement :)

ginty:
Not much different, this is the write! method (I guess you might also do read! while you are at it) - https://github.com/Origen-SDK/origen/blob/master/lib/origen/registers/bit_collection.rb#L509

Something like this should do it:

~~~ruby
yield self if block_given?
# Original code here
~~~

lib/origen/registers/bit_collection.rb
```diff
@@ -509,6 +509,7 @@ def overlay_str
       def write!(value = nil, options = {})
         value, options = nil, value if value.is_a?(Hash)
         write(value, options) if value
+        yield @reg if block_given?
```
ginty: Think these should be yield self and not @reg.

They are effectively the same when operating on a register, but this update does also support the same kind of operation on a subset of bits:

~~~ruby
some_reg.some_bits.write! do |b|
  b[1].write(1)
  b[4].write(1)
end
~~~

By yielding @reg the internal object would be the register and not the subset of bits and any relative position references like above would be off.

chrisnappi: Hmm - that does not work...

chrisnappi: TypeError: no implicit conversion of Symbol into Integer
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/bit_collection.rb:48:in `fetch'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/bit_collection.rb:48:in `block in []'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/bit_collection.rb:47:in `each'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/bit_collection.rb:47:in `[]'
	from (origen):5:in `block in irb_binding'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/bit_collection.rb:512:in `write!'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers/reg.rb:1228:in `method_missing'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/registers.rb:249:in `method_missing'
	from (origen):4
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/commands/interactive.rb:79:in `<module:Origen>'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/commands/interactive.rb:11:in `<top (required)>'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/commands.rb:233:in `require'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/lib/origen/commands.rb:233:in `<top (required)>'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/bin/origen:172:in `require'
	from /proj/mcusoc_kodiak/users/chrisn/origen/origen/bin/origen:172:in `<top (required)>'
	from ./lbin/origen:17:in `load'

ginty: The example as I wrote it above you mean, or something else?

chrisnappi: Using self instead of @reg results in error above

ginty: Just pushed an update that fixes it.
I though `<bit collection>.bits` worked, but it has to be `<reg>.bits`

chrisnappi:
Excellent - thanks!

chrisnappi:
Yikes - this is generating incorrect output (the @reg version worked right).  Doing more testing to determine the issue.

PR #161, opened 2017-10-30
Added in Parameter inheritance between objects and added optional maximum_voltage_rating attribute to Origen::PowerDomain

info-rchitect:
Here is the parameter inheritance usage:

~~~ruby
class DUT
   include Origen::TopLevel

   def initialize
     define_params :default do |params|
       params.tprog = 20
       params.erase.time = 4
       params.erase.pulses = pulses
       params.test.ac.period = 10.ns
       params.test.func = -> { 2 * 2 }
       params.vdd.nom = 1
       params.vdd.min = 0.8
       params.vdd.max = 1.2
    end
    sub_block :ip_with_params, class_name: 'IpWithParams', base_address: 0xDEAD_BEEF
  end
end

    class IpWithParams
      include Origen::Model
      
      def initialize 
        define_params :default, inherit: 'dut.default' do |params|
          params.erase.time = 5
          params.vdd.xmin = 0.7
        end
      end  
    end

  # in parameters_specs.rb
    it 'can pass inheritance between objects' do
      $dut.params.contexts.should == [:default, :ate, :probe, :ft, :set1, :set2, :set3]
      $dut.ip_with_params.params.contexts.should == [:default]
      $dut.params(:default).keys.should == [:tprog, :erase, :test, :vdd]
      $dut.ip_with_params.params(:default).keys.should == [:tprog, :erase, :test, :vdd]
      $dut.params(:default).vdd.keys.should == [:nom, :min, :max]
      $dut.ip_with_params.params(:default).vdd.keys.should == [:nom, :min, :max, :xmin]
      ($dut.ip_with_params.params(:default).vdd.keys - $dut.params(:default).vdd.keys).should == [:xmin]
      $dut.ip_with_params.params(:default).vdd.xmin.should == 0.7
      $dut.params(:default).erase.time.should == 4
      $dut.ip_with_params.params(:default).erase.time.should == 5
    end
~~~

Here is an example of defining a power domain with the optional maximum_voltage_rating attribute:

~~~ruby
   add_power_domain :vdd do |domain|
      domain.description = 'CPU'
      domain.nominal_voltage = 1.0.V
      domain.unit_voltage_range = 0.7.V..1.1.V
      domain.maximum_voltage_rating = 1.50.V
   end
~~~

Here is an example without defining maximum_voltage_rating:

~~~ruby
    add_power_domain :vccsoc do |domain|
      domain.description = 'SoC'
      domain.nominal_voltage = 1.5.V
    end
~~~

If the user specifies the MVR, then the setpoint (if not nil) will be compared against it in the method 'setpoint_ok?'.  @welguisz , can you please check this out in your env?

thx


ginty:
Not sure about the `return_dut_object` method, you can achieve the same thing with:

~~~ruby
eval('dut.ddr')   # => sub-block object
~~~

info-rchitect:
@ginty did we want to introduce eval usage into our code base as the recommended means to solve an issue?  Could cause headaches if people do not know what they are doing.

info-rchitect:
@ginty how does it look now?

ginty:
I am not really a fan of overloading String with business logic which is not concerned with manipulating strings.

I would prefer to see something like this:

~~~ruby
Origen.is_valid_dut_path?('dut.ddr.memc')  # => true
~~~

info-rchitect:
@ginty that works for me, which file should that get added to?

ginty:
If it's just one method, you could just add it to `lib/origen.rb`, if you think there might be more like this you could consider creating a new file/module for this and including it.

info-rchitect:
@ginty OK added.  Where should spec testing be?

info-rchitect:
@ginty are we ready to go here?

lib/origen/parameters.rb
```diff
@@ -100,10 +100,31 @@ def _live_parameter_requested?
 
     private
 
-    def _validate_parameter_set_name(name)
-      unless _parameter_sets.key?(name)
-        puts "Unknown parameter set :#{name} requested for #{self.class}, these are the valid sets:"
-        _parameter_sets.keys.each { |k| puts "  :#{k}" }
+    def _validate_parameter_set_name(expr)
```
ginty: Don't need to change it, but just FYI, since this method is private there is not really any need to make it begin with an underscore.
Normally I would make a method start with an underscore if it was really intended for internal use only but for some reason it could not be made private. I would also mark it as private in the documentation as shown below:

~~~ruby

# @api private
def _an_internal_use_method_but_which_is_not_private
end

private

def a_private_method
end
~~~

ginty: :-) just realized that I may have actually been advising my past self there, but I'll leave it here as a useful note to whoever may see it.

PR #162, opened 2017-10-30
RFC3 - Model import and export

ginty:
This PR implements the full RFC for adding model import and export to Origen core as described in RFC3 - https://github.com/Origen-SDK/rfcs/blob/master/text/0003-model-export-and-import.md

It handles sub-blocks, pins and registers, and at least all of the main attributes of each.

It's possible that some are missing, but it should be easy for the community to further expand `lib/origen/model/exporter.rb` from here.


info-rchitect:
@ginty OK will do

ginty:
Thanks @rlaj, good improvement though 10s is still bad!

Think the other changes planned in this RFC will help with that, or is it something else?

info-rchitect:
@ginty I am testing this now but ran into some issues with the cross-origen code when writing out source files.  stay tuned...

info-rchitect:
@ginty I have been re-working cross-origen to work with this PR.  Now that this PR includes import and export, should I just use that method?  Also, what is the API?

ginty:
@info-rchitect, see the updated overview above.

Cross-origen and any similar importers should be re-worked as follows, where their sole responsibility is for the import side and Origen core will now pick up the exporting to Origen format side of things.

* Importers should parse the 3rd party files and build up an anonymous model using conventional Origen APIs.
* Once they have completed, they should call export on the model like this and give it a name:

~~~ruby
my_model.export("data_from_pintool")
~~~

* This will render everything in the model to the application's vendor directory.
* The application can then consume the imported data by adding an import statement to its model and referring to the same name used for export:

~~~ruby
module MyApp
  class TopLevel
    include Origen::TopLevel

    def initialize(options = {})
      import "data_from_pintool"
    end
  end
end
~~~

* DUT models can call import multiple times to import data from different sources


Origen will have now taken care of structuring the data so that it all lazily loads as the application logic references it. Hopefully, with the end effect that targets comprising thousands of nested sub-blocks containing thousands of registers, should now load very snappily.

Please try and feedback.

info-rchitect:
@ginty I will try right away.  one question: can the export path be configured similarly as the output_directory in the application config?

info-rchitect:
OK first performance shows it takes 55 seconds to export the model.  One show-stopper though is that the export code is not ascertaining the actual app namespace, it must be guessing based on the app name.  My namespace is 'PPEKit' and it made it 'PpeKit'.  I will see if I can find the code and fix it.

info-rchitect:
I went and tried the import method anyways and now have run into what looks like a naming collision. 

~~~shell
[5] pry(#<PPEKit::Product>)> import "#{current_product_name}#{current_product_revision.upcase}"
ArgumentError: wrong number of arguments (given 1, expected 0)
from (pry):5:in `load_regs'
[6] pry(#<PPEKit::Product>)> methods.grep(/import/)
=> [:import, :import=, :import_fuses, :import_test_list, :import_pins, :import_mbist, :import_scan, :import_regs, :cr_import]
~~~

I don't think you made a method called 'import=' did you?  I will start searching for the root cause...

ginty:
Thanks for trying this out so quickly @info-rchitect.

To your points:

* The export path cannot currently be configured except via the name you give to import/export.
* I think the difference between PPEKit and PpeKit should not matter, other than for cosmetic reasons, it should still work fine, though adding the following to your app might make it match:

~~~ruby
require 'active_support/core_ext/string/inflections'
ActiveSupport::Inflector.inflections(:en) do |inflect|
  inflect.acronym 'PPE'
end
~~~

* Yes, there is no `import=`, seems like it must be a collision with something in your app.

info-rchitect:
OK fixed the name collision and now get this error:

~~~shell
[1] pry(#<PPEKit::Product>)> import "#{current_product_name}#{current_product_revision.upcase}"
NameError: uninitialized constant PPEKit::ArielA0
from /home/bcaqueli/.origen/gems/ruby/2.3.0/gems/activesupport-4.2.10/lib/active_support/inflector/methods.rb:263:in `const_get'
~~~

Where should I put the code snippet you suggested @ginty?

ginty:
Put it anywhere in your app before you call export.

We should be able to fix this one in the PR, but that should be the quickest way to proceed for now.

info-rchitect:
@ginty that code snippet did not fix the previous error message I posted.  Should we just bite the bullet and change the namespace detection logic?

ginty:
Try the patch I just pushed

info-rchitect:
OK here is the load times:

~~~shell
[DEBUG]      0.342[0.001]    || PPEKit: Loading IP and registers
[DEBUG]      0.399[0.058]    || PPEKit: Loading of IP and registers complete
~~~

Now don't take this as complaining because it is just the way Ruby works but users should be aware of a couple interactive behaviors that will now exist:

1. The first call to dut.sub_blocks will take quite awhile
2. Users cannot use tab completion for sub_block names anymore until they actually type a sub_block name first.

Overall, very nice PR!

ginty:
@info-rchitect, sounds good, definitely the kind of improvement I expected to see.

Potentially we could fix 2 I think, I'll push an update to try and fix that and you can see if it slows it down much.

For (1), what do you mean by quite a while, is the 10s coming back at that point? I think that should never be the case, but possibly if there is something you are doing that causes all the placeholders to be instantiated at once that would take a long time again.
We should look at fixing that if you can provide more details about what you are doing in case 1.

info-rchitect:
@ginty Behavior #1 takes about 3.5s.

~~~shell
origen(main):003:0> dut.sub_blocks.size
=> 1385
~~~

:)

info-rchitect:
@ginty one minor nitpick.  The path to the model files 'ppe_kit' versus 'ppekit'.  Can you fix that while you are still under the hood?

ginty:
Is it only if you call size on it? i.e. does dut.sub_blocks return instantly?

info-rchitect:
@ginty it is the call to sub_blocks, not the size.  i showed the size to give you some idea of the model size.

ginty:
@info-rchitect, the latest push should hopefully fix the 3.5 seconds thing.

On the auto-complete, I'm actually not sure what to change since I don't see much difference between current and previous implementation in how the accessor methods are defined (which I am assuming is relevant to that).

Can you provide more info on how I can see this one for myself? (never knew there was auto-complete!)

ginty:
@info-rchitect, for the 'ppe_kit' thing, I just added this API which you can call from your application which should help - https://github.com/Origen-SDK/origen/pull/162/files#diff-2b4a21b6169f2e3bf4eb99bc2b4d77a0R105

info-rchitect:
@ginty the new inspect logic made calling sub_blocks instantaneous.

info-rchitect:
@ginty Here is the tab completion feature in the debugger.  I don't know how you would enable this if the singleton method has not been created for the sub_block yet.

~~~shell
origen(main):004:0> dut.engine_
dut.engine_gem_version  dut.engine_versions
origen(main):004:0> dut.engine_bist
=> <SubBlock: engine_bist>
origen(main):005:0> dut.engine_
dut.engine_bist         dut.engine_gem_version  dut.engine_versions
~~~

ginty:
@info-rchitect, never knew you could do that in the console.

Does that patch help?

info-rchitect:
@ginty yep the last commit re-enabled that feature.  Very very handy for devices with thousands of sub_blocks!

PR #163, opened 2017-10-30
Fancy register output

arkohler:
I've changed the register inspect code to add a fancy_output option, default = true.  This option uses extended ASCII characters to draw the tables.
If fancy_output = false, it reverts back to the lower ASCII characters, although I've tweaked the tables a little bit.


info-rchitect:
@arkohler  where is a picture of the new format?

pderouen:
You can see it in the file diffs of approved/inspections.  I'm in favor of pretty output.  :)

info-rchitect:
@arkohler  very nice!

PR #167, opened 2017-11-06
added in String excel_col_index method and added pins method_missing to hit self.meta

info-rchitect:


welguisz:
@info-rchitect Why is this needed in Origen?  Why can't Roo or axlsx handle this?

info-rchitect:
i spose i could go update every gem that deals with Excel but I guess hitting the core library of Origen makes more sense to me, time-wise.

welguisz:
I don't think it needs to be in lib/origen/core_ext.  This seems to be more of a roo function.  I will have to save that if it is import to Origen, it should be placed in lib/roo/excelx/cell.rb in the Origen Core Directory.  The main reason that I suggest that location is that Roo seems to be have no maintainer.

info-rchitect:
@welguisz why would I want  to worry about external gems when a String modifier method like this is common place in Origen already?  Excel, like Verilog, is a 3rd party 'app' that Origen developers need to interact with and there are [at least 6 gems to work with Excel](https://spin.atomicobject.com/2017/03/22/parsing-excel-files-ruby/).  Given that Origen does not, and would never, restrict which gem users could use to interact with Excel it makes more sense to update the core extensions.  While you and I both like roo and axlsx, [rubyXL](https://github.com/weshatheleopard/rubyXL) has a much more active repo so picking and choosing will just make more work in the long run.

welguisz:
@info-rchitect the only thing that I worry about adding things to Origen core is that the code starts to get bloated.  Let's see what @ginty has to say.

info-rchitect:
@welguisz i am fine with getting others opinions but I am not going to update 6 gems or pick and choose our favorite ones.  I am sure the 4 lines of code will easily be a drop in the bucket when we refactor the Specs module (my app side version of Origen::Specs is ~1/5th the size).   If I have to keep it app side that is fine but I would expect anyone parsing Excel files will want this method.

ginty:
I'm not a big fan of overloading string like this, but on the other hand I'm willing to turn a blind eye to a few minor additions if it makes life easier somewhere.

I'd be OK with merging this one.

PR #172, opened 2017-11-11
enhanced model exporter for pins to account for packages

info-rchitect:
here is an example of a pin with package info.

~~~ruby
model.add_pin :tdo, packages: { pcs: { location: 'BB27', dib_assignment: [11816] }, bga: { location: 'BB27', dib_assignment: [11708] } }, meta: { bump_name: 'bp_tdo', cell_type: '1.8V GPIO' }
~~~

Here is interactive mode:

~~~shell
peologin02:ppekit $ origen i
origen(main):001:0> dut.packages
=> [:bga, :pcs]
origen(main):002:0> dut.pins(:tdo).packages
=> {:pcs=>{:location=>"BB27", :dib_assignment=>[11816]}, :bga=>{:location=>"BB27", :dib_assignment=>[11708]} }
~~~
~~~


lib/origen/model/exporter.rb
```diff
@@ -23,6 +23,11 @@ def export(name, options = {})
           export_wrap_with_namespaces(f, options) do |indent|
             f.puts((' ' * indent) + 'def self.extended(model)')
             indent += 2
+            # Write out packages if any
```
ginty: I think this should be wrapped by `if top_level?` , i.e. only write out the top-level packages if the model is the top-level.
This same export method will be applied to all models in the hierarchy.

PR #181, opened 2017-11-30
Fix site_config.gem_install_dir

ginty:
I think this is fixing what was basically a typo, `user_gem_dir` does not exist in the site config and this should have said `gem_install_dir`.

Any attempts to set `Origen.site_config.gem_install_dir` via override in a site_config.yml or via the env variable `ORIGEN_GEM_INSTALL_DIR` were completely broken, now it works.


ginty:
It is definitely working then.

The previous setting does get cached though, so you may need to run fix_my_workspace or start a fresh workspace for it to take effect.

info-rchitect:
@ginty I started a fresh clone and before running any origen commands (e.g. -v) I moved the origen_site_config.yml file into the app root directory.  I then ran origen -v and got the following install directory:

~~~shell
Bundled gems are installed into /home/thisuser/.origen/gems.
Application: 0.1.0
~~~

That is not the correct directory, I defined a different in the origen site config file:

~~~yaml
home_dir: /users/thisuser/origen
~~~

I don't think this is fixed.  Looking at the other related parts of my site config file might help:

~~~yaml
# Define where the gems should be installed
# By default, this will be at <home_dir>/gems but can be overridden here.
#gem_install_dir: /users/thisuser/origen

# ORIGEN STARTUP OPTIONS
# These options define how and where Origen should boot.
# See the guides at advanced/invocations for additional details.

# By default, don't allow user install. This is more for power users to utilize for debug or those who want absolute 
# control over their environment.
user_install_enable: false

# Default to the user's install dir being the .origen in their home directory but allow it to be overridden
#user_install_dir: /users/thisuser/origen
~~~

ginty:
Just tried it with the same setup you describe and it works fine (and that is with Origen 0.27 without this additional fix):

~~~
Bundle complete! 8 Gemfile dependencies, 67 gems now installed.
Bundled gems are installed into /home/stephen/blah/gems.
~~~

Do you have `lbin` or `.bundle` checked in perhaps?

coreyeng:
@ginty,

The user_install_dir was meant for users to be able to move their own gems around themselves. It was meant as a 'user_specific setting' and for tool_repo to be a 'user group setting' and then for 'home_dir' to be the absolute. I purposely got rid of 'gem_install_dir' in the site config and set it to use home_dir instead under the assumption that moving just the gems with the home dir didn't make much sense. I wanted to manage 'gem_install_dir' internally and make the user move it though the aforementioned means.

Now, I definitely should have documented user_install_dir in the invocation guide, but at least to me, gem_install_dir's current behavior was intentional and not broken, but that's just my thought so if you want to add it back in, go ahead.

Thanks.

info-rchitect:
@ginty no we do not have those directories checked in.

image

ginty:
@coreyeng, please fix up this PR as you see fit then.

I am OK with dropping gem_install_dir as a means of changing where the gems live as long as you:

* Remove gem_install_dir from the example site_config
* Put comments on that method telling us that while internally Origen still calls it site_config.gem_install_dir, it is intentional that the value no longer originates from the config variable of the same name.
* Put clear documentation somewhere in the getting started section describing how an average user can easily change where their gems live, (we didn't have this before admittedly with gem_install_dir, but things were much simpler then than they are now). Bottom line, no one should need to go near that advanced page if all they want to do is change their gem dir.
* user_install_dir should probably be mentioned somewhere in the advanced guide, I just assumed that it didn't really exist given that we have all that documentation and it doesn't even mention it.
* We really should be making this backwards compatible with anyone who already has $ORIGEN_GEM_INSTALL_DIR defined

Thanks

pderouen:
As long as I have an environment variable that I can set that will allow me to specify where the gems get installed, I'll be happy.

ginty:
@pderouen, I believe you should be able to set $ORIGEN_HOME_DIR on current and future Origen and it should work, can you try?

@info-rchitect, I am at a loss as to why you see that...

info-rchitect:
@ginty is the file name correct? 'origen_site_config.yml'  Did you check the rest of my site config for correctness?   Do you agree that all I should have to do is un-comment the 'home_dir' variable?

ginty:
@info-rchitect, yes.

I just copied the current origen_site_config.yml into my app root, changed the home_dir variable, ran fix_my_workspace and it installed to the new location.

You don't have another copy in your config dir do you?

info-rchitect:
@ginty config dir is empty.  I didn't run fix my workspace because I was doing this from a fresh clone.  Do I still need to run fix_my_workspace?

ginty:
Should not, but worth trying since I didn't try the fresh clone approach.
They should in principle be the same though.

info-rchitect:
@ginty I followed the [instructions](http://origen-sdk.org/origen/guides/starting/workspace/) and did not get a bin directory created with a file named 'fix_my_workspace'.  Here is what i added to my Gemfile:

~~~ruby
gem 'origen_updater'
~~~

And here is my console session after running 'origen -v':

~~~shell
peologin02:ppekit $ ls bin
ls: bin: No such file or directory
peologin02:ppekit $ ls
config/  environment/  Gemfile.lock  lib/                    pattern/  Rakefile  target/     vendor/
doc/     Gemfile       lbin/         origen_site_config.yml  program/  spec/     templates/
~~~

Can you try a fresh clone as I did and see if i works?

ginty:
@info-rchitect  have you set gem_manage_bundler to false in the site_config?

If you open an `origen i` session and type `require 'origen_updater` it should create it.

The fact you have to do that so suggests that for some reason Origen is not in control of the load process properly, which is probably related to this other problem too.

coreyeng:
@ginty, I was intentionally blocking user's moving their gems directories without some knowledge of how Origen boots and what exactly they were changing. Even here (NXP), I think most casual users still don't really understand how gems are populated/stored. We can put as much documentation online as we want and link to all the Ruby gems pages we can, but I still get questions about it.

I think what @info-rchitect really wants is a way to change everybody's default gem path to be somewhere else. In that case, I think home_dir should be changed at the root site config (along with the custom gem server/ldap stuff) and I should make home_dir auto-append some type of /<who_am_i>/.origen. This way, all user's would see their home_dir's move. This would move the gems, but also move where the their global Gemfile is, where the .session is stored, etc.

Moving just the user_gem_dir would move where the gems are installed while keeping the global Gemfile and session in <home_dir>/.origen. 

I did this to try and it make as customizable as possible, but maybe now it's just confusing and outstripping its usefulness?

Anyway, in addition to some the bullets you've listed, I'd add fixing home_dir variable to just be a root path from the site config, so that if @info-rchitect sets /users/ instead of ~/, he'll actually get /users/<whoami>/.origen instead of ~/.origen. I'll just a case where if the home_dir starts with ~/, it will resolve to just ~/.origen. I think this will give @info-rchitect the true desired behavior, while all the new/casual users need not be concerned with where/how their home_dirs are getting set.

@info-rchitect, please correct me if any of that is wrong.

I would also put in some command or append to origen -v to print the current home_dir. So if users are expecting a ~/.origen and don't see it, they can easily learn where their origen home directory is without having to worry about where/how its getting set.

pderouen:
@ginty, yes, ORIGEN_HOME_DIR gets the job done.

info-rchitect:
@ginty  here is my sit config setting:

~~~yaml
# When true Origen will configure and run Bundler in a consistent way for every
# user (recommended)
gem_manage_bundler: true
~~~

ginty:
Ok @coreyeng, see what you think of that, adding `user_gem_dir` back along with `gem_install_dir`

info-rchitect:
will there be a docs update with this PR?

ginty:
Thanks for the doc update @coreyeng - http://origen-sdk.org/origen/guides/starting/workspace/#Workspace_Size_Considerations

PR #182, opened 2017-11-30
Fixed compiler bug if template files contained multiple '.'

info-rchitect:
files such as 'shell.lev.dpspins.erb' would fail before.


info-rchitect:
any issues with a merge today?

PR #187, opened 2017-12-05
Updated gems to include greater than 1.7.2 for nokogiri

welguisz:
When I upgraded to 2.4, nokogiri would not install.  I made a small change to get my app to install.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -16,7 +16,7 @@ PATH
       log4r (~> 1.1.10, ~> 1.1)
       nanoc (~> 3.7)
       net-ldap (~> 0.13)
-      nokogiri (= 1.7.2)
+      nokogiri (>= 1.7.2)
       pry (~> 0.10)
       rake (~> 10)
       rspec (~> 3)
@@ -176,4 +176,4 @@ DEPENDENCIES
   ripper-tags
 
 BUNDLED WITH
-   1.15.4
+   1.16.0
--- a/origen.gemspec
+++ b/origen.gemspec
@@ -32,7 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "net-ldap", "~>0.13"
   spec.add_runtime_dependency "httparty", "~>0.13"
   spec.add_runtime_dependency "bundler", "~>1.7"
-  spec.add_runtime_dependency "nokogiri", "1.7.2"
+  spec.add_runtime_dependency "nokogiri", ">= 1.7.2"
   spec.add_runtime_dependency "rspec", "~>3"
   spec.add_runtime_dependency "rspec-legacy_formatters", "~>1"
   spec.add_runtime_dependency "thor", "~>0.19"
--- a/origen.gemspec
+++ b/origen.gemspec
@@ -32,6 +32,9 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "net-ldap", "~>0.13"
   spec.add_runtime_dependency "httparty", "~>0.13"
   spec.add_runtime_dependency "bundler", "~>1.7"
+  # @welguisz ... If your application depends on 1.7.2 and doesn't work with 
+  # anything later than 1.7.2, add the following line into your Gemfile
+  # gem 'nokogiri', '1.7.2'
   spec.add_runtime_dependency "nokogiri", ">= 1.7.2"
   spec.add_runtime_dependency "rspec", "~>3"
   spec.add_runtime_dependency "rspec-legacy_formatters", "~>1"


welguisz:
I was doing a bundle update for origen_app_generators
~~~term
%\origen_app_generators\> bundle update
Fetching gem metadata from https://rubygems.org/............
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    origen_app_generators was resolved to 1.1.3, which depends on
      origen (>= 0.23.0) was resolved to 0.28.1, which depends on
        nokogiri (= 1.7.2) was resolved to 1.7.2, which depends on
          ruby (< 2.4) x86-mingw32

    origen_app_generators was resolved to 1.1.3, which depends on
      origen (>= 0.23.0) was resolved to 0.28.1, which depends on
        nokogiri (= 1.7.2) was resolved to 1.7.2, which depends on
          ruby (>= 2.1.0)

    origen_app_generators was resolved to 1.1.3, which depends on
      ruby (>= 1.9.3)
~~~

When I look at the [Nokogiri Change Log](https://github.com/sparklemotion/nokogiri/blob/master/CHANGELOG.md), I don't see anything that says ruby has to be less than 2.4.

My thoughts if you need to lock it to 1.7.2, the application GemFile should have the following:

gem 'nokogiri', '< 1.8.0'

This will resolve to nokogiri, version 1.7.2

ginty:
Moving to `< 1.8.0` is not really a solution either, since if 1.7.3 gets released it could break many peoples environments.

So it seems to be a problem with running Ruby 2.4 on Windows, can you not use an earlier version? The current Windows install guide says to use 2.2.4.

welguisz:
@ginty ... I am getting that error in Linux, not Windows.  If a person needs to a specific appliation, they can place it in their GemFile, e.g.

~~~ruby
gem 'nokogiri', '1.7.2'
~~~

coreyeng:
The webpage doesn't say anything about having that version requirement (https://rubygems.org/gems/nokogiri/versions/1.7.2)... and neither does the gemspec actually. That's seems odd. All it has is: 

<code>s.required_ruby_version = Gem::Requirement.new(">= 2.1.0".freeze)</code>

But I don't see how you'd be getting <code>x86-mingw32</code> if you were running on Linux. Seems like a larger environment issue to me, at least.

coreyeng:
I wonder if just the Gemfile.lock needs rebuilding when moved from Windows to Linux.

ginty:
@welguisz I think it must be something like @coreyeng mentions.

Recommend you try to fix this via the fix_my_workspace script - http://origen-sdk.org/origen/guides/starting/workspace/#Fix_My_Workspace

I just tried it under Ruby 2.4 and no issues:

~~~
nxa21353@awv120484.nxdi.us-cdc01$ ruby -v
ruby 2.4.2p198 (2017-09-14 revision 59899) [x86_64-linux]
nxa21353@awv120484.nxdi.us-cdc01$ origen i
origen(main):001:0> require 'nokogiri'
=> false
origen(main):003:0> Nokogiri::VERSION
=> "1.7.2"
~~~

welguisz:
~~~
...origenSDK/origen_app_generators 220$ ./bin/fix_my_workspace 
Fetching gem metadata from https://rubygems.org/............
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    origen_app_generators was resolved to 1.1.3, which depends on
      origen (>= 0.23.0) was resolved to 0.28.1, which depends on
        nokogiri (= 1.7.2) was resolved to 1.7.2, which depends on
          ruby (< 2.4) x86-mingw32

    origen_app_generators was resolved to 1.1.3, which depends on
      ruby (>= 1.9.3)
Fetching gem metadata from https://rubygems.org/............
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
Bundler could not find compatible versions for gem "ruby":
  In Gemfile:
    ruby

    origen_app_generators was resolved to 1.1.3, which depends on
      origen (>= 0.23.0) was resolved to 0.28.1, which depends on
        nokogiri (= 1.7.2) was resolved to 1.7.2, which depends on
          ruby (< 2.4) x86-mingw32

    origen_app_generators was resolved to 1.1.3, which depends on
      ruby (>= 1.9.3)
~~~

welguisz:
The issue was that the Gemfile.lock was bad.  It was checked in with Windows dependencies.  Deleted Gemfile.lock and regenerated from Gemfile.  The new Gemfile.lock works.

PR #188, opened 2017-12-05
updated model importer to allow user specified path

info-rchitect:
Added this for the use case where users copy files from default vendor path to a common disk for sharing across users.

### DIFF
--- a/lib/origen/application/configuration.rb
+++ b/lib/origen/application/configuration.rb
@@ -19,7 +19,7 @@ module Origen
                     :web_domain,
                     :strict_errors, :unmanaged_dirs, :unmanaged_files, :remotes,
                     :external_app_dirs, :lint_test, :shared, :yammer_group, :rc_url, :rc_workflow,
-                    :user_aliases, :release_externally, :gem_name, :disqus_shortname
+                    :user_aliases, :release_externally, :gem_name, :disqus_shortname, :model_export_dir, :model_import_dir
 
       # Mark any attributes that are likely to depend on properties of the target here,
       # this will raise an error if they are ever accessed before the target has been
@@ -32,7 +32,7 @@ module Origen
         :pattern_header, :release_directory, :pattern_name_translator, :pattern_directory, :pattern_output_directory,
         :proceed_with_pattern, :test_program_output_directory, :test_program_source_directory,
         :test_program_template_directory, :referenced_pattern_list, :program_prefix, :web_directory,
-        :web_domain
+        :web_domain, :model_import_dir, :model_export_dir
       ]
 
       # Any attributes that want to accept a block, but not necessarily require the target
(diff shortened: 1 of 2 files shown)


info-rchitect:
@ginty doesn't the name used in the model exporter 'namespace' it?

ginty:
It namespaces the code, but not the file name. All file names that are added to Ruby's global load path should be name spaced. The default location here is to store in vendor/lib which means it will be in the global load path.

If you supply a path to store it somewhere outside of the lib directories, then this concern goes away.

info-rchitect:
@ginty as I continue to validate this PR I see that the vendor relative path is hard-coded in a couple spots.  Could we allow apps to set an 'export_dir' that defaults to the vendor path?  This would allow apps to override this in the config file.

ginty:
I don't think I really like the direction this is going, though I appreciate why you've gone down that route.

Consider the case where a plugin might contain some models which it locally imports from its `vendor/lib` directory (included in its gem package).
By globally overriding the import/export path like this at app-level it will break the plugin's import for its model.

The `export_path` method is really the key to this, and if you supply a path to `export` or `import` it should set a class variable, e.g. `@@export_path` which would globally override it for that particular export/import and then restore it at the end.

This is the API we want to support I think:

~~~ruby
# Default's to the vendor dir as today
export :blah   
import :blah

# Override for specific imports
export :blah, file_path: "/my/data/files"
import :blah, file_path: "/my/data/files"

# Batch override
with_file_path "/my/data/files" do
  import :blah
  import :blah2
  import :blah3
end
~~~

info-rchitect:
@ginty i don't mind the API you suggest but for us we may not (and on the first product did not) know ahead of time where those Ruby files will be placed.  I know I cannot guarantee we will ever become that disciplined.  I still don't understand the plugin issue as well as I wish I did.

ginty:
Don't see the relevance about when you know about the location, both approaches let you change it whenever you want and my recommendation is more flexible since you can control it on a per-import basis.

I know you always create top-level models that contain everything, which is fine and we do that too sometimes, but other times we model more at the IP-level. When we do that, the IP block model lives in a plugin and the owner of it could write it by hand, or they could import some design IP which they export to the plugin's vendor dir.

If that plugin then gets used in a top-level app where the import dir is globally pointed to somewhere else, then the plugin's import will break.

info-rchitect:
@ginty the relevance is that the new API hard-codes an absolute path versus using a relative path which makes the files usable anywhere at any time.  Or did I mistake the API you suggested?

ginty:
File path was a bad name, please disregard and call it something better, it means the export_dir.

I almost called in export_dir, but I decided not to since I thought maybe it looked a bit weird specifying the export_dir when calling import, but maybe not on reflection.

Anyway the argument just means the base directory, `Origen.root/vendor/lib` be default and here adding the ability to override that to be some other dir.

info-rchitect:
@ginty gotcha, thanks for the explanation.  I will probably just drop this PR and start over.

Issue #183, opened 2017-12-01
Origen home_dir site config variable


coreyeng:
Following the discussion from PR #181 I'd like to open the topic of having the home_dir site config auto-append some user specifics.

My intention of this was to adequately move the entire 'user install directory' to somewhere else. This was intended to be a global site-config setting, essentially fully switching ~/.origen to home_dir. Currently, it will accomplish this and can be done for a single user effectively, but it can't be set by a origen install manager as originally intended, as then everyone will end up with home_dir and start stepping on each other.

My proposal is to auto-append, the username to home_dir, rendering home_dir as just a root path. For example:

- home_dir: ~/ #=> (default) Creates Origen user installation at ~/.origen
- home_dir: /proj/origen_workspaces #=> (set in site config for example) Create user installations at /projge/origen_workspaces/<whoami>/.origen

I could add another site variable (maybe disable_home_dir_appendix) that would disable this, but in that case I had the user_install_dir, user_gem_dir, user_install_enable site configs to handle this.

Please let me know what your thoughts are on this. I want to balance customization, usefulness, and clarity. I always use the combo user_install_dir, user_gem_dir, and user_install_enable, so I'm not sure how others are looking to use this and what results they are expecting.

Once I've gathered feedback I'll start implementing.

@ginty, per your notes from PR #181 

- Remove gem_install_dir from the example site_config
Corey: I will do this for this issue. I also added a deprecation note in the docs saying this will be overruled if 'user_gem_dir' is also present.
- Put comments on that method telling us that while internally Origen still calls it site_config.gem_install_dir, it is intentional that the value no longer originates from the config variable of the same name.
Corey: I'll redo the comments for clarity in site_config.rb.
- Put clear documentation somewhere in the getting started section describing how an average user can easily change where their gems live, (we didn't have this before admittedly with gem_install_dir, but things were much simpler then than they are now). Bottom line, no one should need to go near that advanced page if all they want to do is change their gem dir.
Corey: Updated the docs for PR #181 but will make a second pass though those docs with this issue.
- user_install_dir should probably be mentioned somewhere in the advanced guide, I just assumed that it didn't really exist given that we have all that documentation and it doesn't even mention it.
Corey: Will do for this issue.
- We really should be making this backwards compatible with anyone who already has $ORIGEN_GEM_INSTALL_DIR defined
Corey: Completed by you in PR #181

ginty:
Hi @coreyeng, thanks for following up on this.

[...]

PR #207, opened 2018-01-04
Express global commands

coreyeng:
The TL;DR of this: Origen can now register plugins installed at the system level and can run global commands defined there. Also, this includes the enhancements for issue #183 and some methods to dynamically mess with the site config.

Hi,

This is a PR to add a requested feature which allows for global commands to be run when gems are installed at the system level.

Previously, in order to run global commands, you had to setup a TR or user install. TR users could reference a TR install, but that still required some sort of setup (usually a shell script to setup environment variables). With this PR, gems installed at the system level (using gem install), will be registered as Origen plugins, and are capable of adding global commands.

For example, to run a global command in the gem <code>te_utils</code>, the Origen system install manager can simply do: <code>gem install te_utils</code> and run any global commands provided by <code>te_utils</code>. This will go through the same process as the standard boot up routine, so everything required by the global commands will still be available.

Bundler is bypassed in this case. Under the hood, this is just a very simple check to see if Bundler is being used (if it is, then this is an application/tr/user invocation, otherwise it’s a system level invocation), and if not then it will loop through all the registered gems in the system, looking for gems with a dependency on Origen, and load those applications, registering them as plugins.

This opens up two caveats that I see:

- Development and debug is a bit more involved. As far as I can see, everything Bundler does Gem can do as well (which makes sense) just is a bit more involved, and potentially a bit more error prone. For example, developing a gem locally and installing it at the system level is a bit more involved instead of just using the <code>:path</code> option in Bundler. However, all of this should work out to where there is a 1-to-1 correlation between using the system install and the user install so that running from a system install resolves the same as running the user install.
- In normal circumstance, using <code>gem install</code> will resolve and install all dependencies. So, we can safely <code>require</code> origen plugins installed using gem in the same way as using Bundler. However, I think that hacking at the gems install is more dangerous than hacking at Bundler's install (less checks, but maybe I'm wrong), so trying to cut corners during development could cause issues. So, I recommend developing global commands using 1).

This is not so much a caveat, but a design decision: if Bundler is used (we're running from an application/user/TR invocation) then this feature is skipped. This is to remove Origen automagically loading gems that may not be included in the Gemfile. This can be problematic for applications being used across company systems, across companies, or across Windows users. So, to take out the possibility of random 'gem not defined' type issues, the same gem must be required by the Gemfiles and the system, if it is to be used by both invocations.

Docs were updated, but I'm not sure how specs tests could be added for this. I'm not familiar with how the specs are actually run on Git's server or if specs for this would even be possible.

Additionally, this includes fixes/enhancements for issues #183. Docs and specs tests also updated accordingly. I think I hit on all the bullets Ginty had. The only one that I purposely didn't do was having <code>user_gem_dir</code> in the advanced section. But I covered this pretty well I think in the <code>workspace</code> section and Ginty had mentioned that if we just want to move the <code>user_gem_dir</code> the advanced pages shouldn't be needed. The advanced pages are more about setting up user and TR installs anyway and putting the same content that is covered in the <code>workspace</code> section seemed repetitive and out of place to me.

<strong>I changed up the site config stuff quite a bit. So please do review. I don't think it will substantially change anything though but is worth mentioning.</strong>

Lastly, when prototyping for #183, I added some dynamic stuff to get/add/remove values to/from the site config. Seemed like a waste to just throw those away so I added some specs for those too and updated the site config docs. I have a warning though that basically says its use is limited to after Origen has already booted.

### DIFF
--- a/lib/origen/commands_global.rb
+++ b/lib/origen/commands_global.rb
@@ -23,6 +23,19 @@ if ENV['BUNDLE_GEMFILE']
   Bundler.require(:development)
   Bundler.require(:runtime)
   Bundler.require(:default)
+else
+  # If we're not running from a Bundler build, which we aren't here,
+  # get a list of installed system gems. Go through this list finding and get any that has a dependency
+  # on Origen. If so, assume that gem is a plugin.
+  # For all plugins, require it to register as a plugin.
+  # The global handler below will take it from there.
+  Gem::Specification.each do |gem|
+    gem.dependencies.each do |d|
+      if d.name == 'origen'
+        require gem.name
+      end
+    end
+  end
 end
 
 # Load the global app and an empty target, this helps to ensure that all of Origen's functionality
--- a/lib/origen/site_config.rb
+++ b/lib/origen/site_config.rb
@@ -16,16 +16,16 @@ module Origen
       else
         path = eval_path(find_val('user_gem_dir') || find_val('gem_install_dir') || home_dir)
       end
-      
+
       append = find_val('append_gems')
-      append = 'gems' if (append == true || append.nil?)
-      
+      append = 'gems' if append == true || append.nil?
+
       if append
-        if !path.end_with?(append)
+        unless path.end_with?(append)
           path = File.join(path, append)
         end
       end
-      path      
+      path
     end
     alias_method :user_gem_dir, :gem_install_dir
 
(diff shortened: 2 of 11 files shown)


coreyeng:
Also, there's a kind of annoying formatting issue when using the code tags. I don't think it likes the backslashes. What's currently in there will generate as (setup as Ruby):

image

Using no formatting looks like:

image

Even though the formatting is off, I think the former still looks better. But... I could be convinced otherwise. Any opinions?

info-rchitect:
@coreyeng to be honest I just haven't used many custom commands so I am more just learning here, don't have any feedback yet.  But thanks for including me!

coreyeng:
Hi @info-rchitect, no one has really. Even myself and Ginty haven't been using more than a few. But, this also includes the changes to Issue #183, which was opened in light of the home_dir/gems_dir and all that being moved around. I think you were having your engineers using the ORIGEN_GEM_INSTALL_DIR environment variable override. Now, you should be able to place something like home_dir: /larger_drive/~and have the ~ resolve to their username. So, for me, that would become: /larger_drive/coreyeng/.origen and move everything.

Then I made a bunch of specs and updated the docs which should help give show that we're supporting it. So, you could share the updated docs with your engineers. May help the sell.

ginty:
@coreyeng, if you want to include Ruby formatted snippets in here, just do:

~~~
~~~ruby
# Your Ruby code here
\~~~
~~~

**The backslash on the last line should not be used, but I couldn't get it to print an example code block here without it!**

ginty:
Sorry, @coreyeng, just realized that you were talking about formatting in the Origen web pages rather than here.
I guess your example looks a bit funky since the top part of the block is not valid Ruby, it is YML, which I think the syntax highlighter would handle OK if you split it into a ~~~yaml block and a ~~~ruby block.

For reference, this is the code highlighter that is used when generating Origen web pages and any language listed there should be handled correctly: http://coderay.rubychan.de/

coreyeng:
Thanks @ginty! I thought I tried using yaml but it still looked funky. Maybe I had a typo. I might have used yml instead by mistake. I didn't actually know what that tag was called, so that link is helpful. Thanks again!

I haven't tried to use the global commands on Linux, since I can't update the system Ruby, but they worked fine for me on Windows. Should be fine on Linux though.

PR #209, opened 2018-01-11
Added printing the name of the app/plugin that included the remote.

coreyeng:
This was a change I've had in my workspace for a while. Very simple change to just print the name of the requiring plugin for remotes. I added this to help debug a startup issue when I knew what the remote was and why it wasn't working (URL changed) but had no idea which plugin was actually including it to go fix it. For example, the output will now look like:

image

### DIFF
--- a/lib/origen/remote_manager.rb
+++ b/lib/origen/remote_manager.rb
@@ -22,9 +22,9 @@ module Origen
           Origen.log.info 'The following remotes need to be updated, this will now happen automatically:'
           dirty_remotes.each do |name, remote|
             if remote[:path]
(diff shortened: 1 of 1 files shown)


info-rchitect:
@coreyeng could the same remote be required by multiple plugins?

coreyeng:
If the same remote is in multiple times it should only 'download' once. So in that case, it would warn the first time its encountered. If it tries to grab a different version of the remote then it will be run again. The dialog would look the same but the git/designsync/whichever command it prints when it says "This command failed" will change.

I didn't actually test that but I think that's how it will work. In any case, it shouldn't cause additional confusion, or worse, break something.

PR #210, opened 2018-01-11
Update to windows install docs changing recommended versions of Ruby …

coreyeng:
This is another update I've hung onto for a while (for no reason in particular) that I addressed for some of our Windows users. The current install guide actually does not work for us. We get issues with some gems (the ones trying to build libraries, like JSON, nokogiri, etc) not installing due to <code>Symlink</code> issues.

We found that we can get around this by using RubyGems 2.2.3 (what the original Windows install guide had) or by manually installing the gems with some options to not build the libraries.

[However, with Ruby 2.2 support ending in March](https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/) (see the last paragraph on that page), I've been recommending to our users to just install Ruby 2.3 and Ruby Gems 2.7.3, and this has worked well for us. Since the specs that we run on our Linux environment (at least here at NXP) use Ruby 2.3, I think we should update the Windows install guide to these new versions. I've been using them for a while now and haven't seen any issues. I actually prototyped #207 with this install.

I also added a box that says don't stress too much about the versions of Ruby/RubyGems and described a bit of our struggle. I hope it doesn't come off as hating on RubyGems, because I actually love RubyGems, but shows to try different versions and use best judgement if running into gem installation issues.

### DIFF
--- a/templates/web/guides/starting/installwin.md.erb
+++ b/templates/web/guides/starting/installwin.md.erb
@@ -82,8 +82,10 @@ choco install consoleZ -y --force
 
 ### 3. Install Ruby
 
+#### 3.1 Install the Ruby Executable
+
 ~~~text
-choco install ruby -y --version=2.2.4 --force
+choco install ruby -y --version=2.3 --force
 choco install ruby2.devkit -y --force
 ~~~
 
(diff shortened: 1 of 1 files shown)


priyavadan:
LGTM!

PR #226, opened 2018-02-19
updated clock API to have min/max methods based on freq range.  Creat…

info-rchitect:
Eventually Origen::Specs would incorporate this module.  Why would someone use Limits by itself?  Origen::Specs is designed to model a four dimensional specification universe that is more than what many folks need.  This PR also includes the enhancement of allowing users to create limit sets versus just one static limit set.  Users can reference other objects using symbols in formulas with the following scope hierarchy:

1. Another limit set
2. Power domains
3. Clocks

I think eventually the user will be able to pass in any object path (similar to Parameters) as an option to change where to look for symbol references.  Here are some examples of adding in limits to tests, for each of the 3 cases above.

~~~ruby
    add_test :pin_leakage do |t|
      t.description = 'pin leakage, pin to GND and pin to RAIL'
      t.conditions  = [:minvdd, :maxvdd]
      t.platforms   = [:v93k]
    end.add_limits(:eds, min: (EDS_PIN_LKG_SPEC * -1), max: EDS_PIN_LKG_SPEC, )
    tests(:pin_leakage).add_limits(:ws, min: -100.nA, max: 100.nA)
    tests(:pin_leakage).add_limits(:ft, min: -500.nA, max: 500.nA)
    add_test :voh do |t|
      t.description = 'I/O voltage output high'
      t.conditions  = [:minvdd, :maxvdd]
      t.platforms   = [:v93k]
    end.add_limits(:eds, min: ":vddio * 0.7" )
    tests(:voh).add_limits(:ws, min: ":vddio * 0.55")
    tests(:voh).add_limits(:ft, min: ":eds - 10.mV")
    add_test :fmin do |t|
      t.description = 'Frequency Min'
      t.conditions  = [:minvdd, :maxvdd]
      t.platforms   = [:v93k]
    end.add_limits(:eds, min: :core_clk )
    tests(:fmin).add_limits(:ws, min: :eds)
    tests(:fmin).add_limits(:ft, min: ":eds + 50.Mhz")
~~~

Hope this explains it better @ginty

### DIFF
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -68,6 +68,7 @@ unless defined? RGen::ORIGENTRANSITION
     autoload :Clocks,            'origen/clocks'
     autoload :Value,             'origen/value'
     autoload :OrgFile,           'origen/org_file'
+    autoload :Limits,            'origen/limits'
 
     attr_reader :switch_user
 
--- a/lib/origen/clocks/clock.rb
+++ b/lib/origen/clocks/clock.rb
@@ -78,6 +78,24 @@ module Origen
       end
       alias_method :target, :freq_target
 
+      # min method
+      def min
+        if @freq_range == :fixed
+          @freq_target
+        else
+          @freq_range.first
+        end
+      end
+
+      # max method
+      def max
+        if @freq_range == :fixed
+          @freq_target
+        else
+          @freq_range.last
+        end
+      end
+
       # Check if the clock users are defined anywhere in the DUT
       def users_defined?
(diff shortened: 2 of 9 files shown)


ginty:
I know that the update contains specs, but could you update the overview here with some examples of how this will be used?

info-rchitect:
@welguisz any issues with this PR?

info-rchitect:
OK, no objections on the fifth day at noon will result in a merge.

PR #237, opened 2018-02-27
on_mode_changed callback added

info-rchitect:
First discussed [here](https://stackoverflow.com/questions/48811004/changing-object-data-using-with-mode-method).

~~~ruby
  class MyModel
    include Origen::Model
    attr_reader :my_version
    attr_accessor :mode_changed_count
    def initialize
      @my_version = version
       add_mode :mode1
      @mode_changed_count = 0
    end
    
    def on_mode_changed(options)
      @mode_changed_count += 1
    end
  end

    it 'on_mode_changed callback works' do
      m = MyModel.new
      m.modes.should == [:mode1]
      m.mode.should == nil
      m.mode_changed_count.should == 0
      m.mode = :mode1
      m.mode_changed_count.should == 1
    end
~~~

### DIFF
--- a/templates/web/guides/misc/callbacks.md.erb
+++ b/templates/web/guides/misc/callbacks.md.erb
@@ -144,13 +144,13 @@ There now follows a listing of all available callbacks, these have been split up
 
 * [Environment Setup](#environment_setup)
 * [Environment Teardown](#environment_teardown)
+* [Mode Changed](#mode_changed)
 * [Generation](#generation)
 * [Reset](#reset)
 * [Pattern Generation](#pattern_generation)
 * [Program Generation](#program_generation)
 * [Web Compiler](#web_compiler)
 * [Release Process](#release_process)
-* [Intializing Models](#initializing_models)
 
 <a class="anchor" name="environment_setup"></a>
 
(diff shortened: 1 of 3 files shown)


templates/web/guides/misc/callbacks.md.erb
```diff
@@ -266,6 +267,15 @@ guaranteed to be called even if the thread crashes.
 An example use cases is to ensure that any temporary files are deleted.
 
 
+<a class="anchor" name="mode_changed"></a>
+
+### Mode Changed
+
+#### on_mode_changed
+
```
ginty: Internal links should be created using this helper:

<%= path "api/Origen/Model.html#current_mode=-instance_method" %>

That means they will work when generating offline, and generally if the root url changes, e.g. if we move to https in future.

templates/web/guides/misc/callbacks.md.erb
```diff
@@ -266,6 +267,15 @@ guaranteed to be called even if the thread crashes.
 An example use cases is to ensure that any temporary files are deleted.
 
 
+<a class="anchor" name="mode_changed"></a>
+
+### Mode Changed
+
+#### on_mode_changed
```
ginty: Should reflect that this takes an options hash.

Should also note here that the new mode will be passed in via options[:mode].

templates/web/guides/misc/callbacks.md.erb
info-rchitect: where does the text go?

ginty: The above just replaces http://...., so it would be:

`[changes its current mode](<%= ... %>).`

info-rchitect:
thx @ginty.  This will allow us to model the device modes without having to duplicate a bunch of clock information inside the ChipMode object.  The ChipMode object is handy for storing things you haven't or don't want to really model but if an object model exists then this callback seems like a better option.

info-rchitect:
completed [here](https://github.com/Origen-SDK/origen/pull/237)

Issue #229, opened 2018-02-20
Revision control adding 'v' to remote versions


info-rchitect:
Hi,

The Origen revision control module (in this case Git) seems to be pre-pending a 'v' to the remote version.  I tagged my remote repo as '0.1.0' and it failed.  Here is my remote config:

~~~ruby
  config.remotes = [
    {
      dir: 'product',
      vault: 'ssh://git@stash.mycompany.com:7999/myproduct/origen.git',
      version: '0.1.0'
    }
  ]
~~~

Here is the error:

~~~shell
peologin02:ppekit $ rm -rf ../.ppekit_remotes_DO_NOT_HAND_MODIFY/
peologin02:ppekit $ origen i
[INFO]       0.017[0.017]    || *******************************************************************************
[INFO]       0.017[0.000]    || The following remotes need to be updated, this will now happen automatically:
[INFO]       0.018[0.001]    ||   product - 0.1.0 (included by ppekit)
[INFO]       0.018[0.000]    ||
[INFO]       0.056[0.000]    || git init
[INFO]       0.056[0.000]    ||
[INFO]       1.340[1.284]    || Initialized empty Git repository in /users/user/origen/.ppekit_remotes_DO_NOT_HAND_MODIFY/product/.git/
[INFO]       1.361[0.021]    || git remote add origin ssh://git@stash.mycompany.com:7999/myproduct/origen.git
[INFO]       1.361[0.000]    ||
[INFO]       1.389[0.028]    || git fetch
[INFO]       1.389[0.000]    ||
[INFO]       5.520[4.131]    || From ssh://stash.mycompany.com:7999/myproduct/origen
[INFO]       5.629[0.095]    || * [new branch]      master     -> origin/master
[INFO]       5.637[0.007]    || * [new tag]         0.1.0      -> 0.1.0
[INFO]       5.892[0.255]    || git reset --hard v0.1.0
[INFO]       5.892[0.000]    ||
[INFO]       5.912[0.020]    || fatal: ambiguous argument 'v0.1.0': unknown revision or path not in the working tree.
[INFO]       5.912[0.000]    || Use '--' to separate paths from revisions, like this:
[INFO]       5.913[0.000]    || 'git <command> [<revision>...] -- [<file>...]'

COMPLETE CALL STACK
-------------------
When updating remotes for ppekit: This command failed: 'git reset --hard v0.1.0'
~~~

[...]

PR #243, opened 2018-03-08
Enhancement for remotes and new Perforce RC class

info-rchitect:
This PR adds in a very stripped down Perforce revision control API that currently just handles downloading a file in read-only mode, which works perfectly for Origen remotes.

The remotes manager fixes/enhances two things:

* This [issue](https://github.com/Origen-SDK/origen/issues/229)
* This [discussion](https://stackoverflow.com/questions/48868770/can-callbacks-be-used-to-configure-remotes) about the inability to redefine Origen remotes outside the application file and force them to download

Here is an example of using the new remote config key 'tag', which will not get a 'v' added to it if it looks like a semantic version. It also shows off the Perforce syntax for remotes.

~~~ruby
    Origen.config.remotes do
      [
        {
          dir: "mydir",
          vault: "p4://myserver:myport//path/to/my/remote/file/myfile.xlsx",
          tag: '0.1.0'
        }    
      ]
    end
~~~

There is also a new resolve_remotes! method that forces remotes to download, without having to check if they are dirty.


lib/origen/remote_manager.rb
```diff
@@ -119,8 +119,46 @@ def symlink?(path)
...
 
+    def resolve_remotes!
+      resolve_remotes
+      process_remotes
+    end
+
     private
 
+    # Process each remote
+    def process_remotes
```
ginty: Could you comment on why this method is being added, is there nothing like this in here already?

lib/origen/remote_manager.rb
```diff
@@ -243,6 +281,11 @@ def remotes
     #   * If multiple versions of the same remote are found the most
     #     recent one wins.
     def add_remote(new)
+      # Cannot have both a tag and a version defined for a remote
+      if ([:tag, :version] - new.keys).empty?
+        Origen.log.error('Cannot define both a tag and a version for a remote!')
```
ginty: This should be:

~~~ruby
fail "Cannot..."
~~~

i.e.  better to show the user the stack trace if going to kill the execution

lib/origen/revision_control/base.rb
```diff
@@ -192,6 +195,11 @@ def git?
         is_a?(Git) # :-)
       end
 
+      # Returns true if the revision controller object uses Git
+      def p4?
+        is_a?(Perforce) # :-)
```
ginty: The joke only works for Git!

ginty:
Seems like a good update to me, just surprised it was necessary to add the process_remotes method, thought I haven't looked to see if there were alternative implementation possibilities.

lib/origen/remote_manager.rb
info-rchitect: @ginty When I tried to use the require! method, it didn't work because it was looking only for existing remotes:

~~~ruby
def require!
      unless required?
        while updates_required?
          Origen.log.info '*******************************************************************************'
          Origen.log.info 'The following remotes need to be updated, this will now happen automatically:'
          dirty_remotes.each do |name, remote|
~~~

Because my remotes are not defined in the application.rb file, they never get created, and then cannot be dirty_remotes, which get re-assessed.  I just need them to be created.

ginty: OK, I started trying to working out an alternative implementation which didn't introduce a new method but couldn't see an easy way either

lib/origen/remote_manager.rb
```diff
@@ -119,8 +119,46 @@ def symlink?(path)
       end
     end
 
+    def resolve_remotes!
```
ginty: Can you add some # comments here as to what the method does and when/why you would call it?

After that, this is good to go from my perspective.

Thanks!

info-rchitect:
@ginty are we good to go?

info-rchitect:
@ginty are we good to go even with coverage decrease?  it is to be expected because RC doesn't have specs right?

ginty:
Something strange is going on with the new doc files, looks like maybe you've copied some guide pages to the top-level dir then checked them in?

info-rchitect:
@ginty how about now?

PR #244, opened 2018-03-09
Exporter fix and support parallel interface reg sync

ginty:
Some minor tweaks to better handle `namespace: nil` when importing, based on experience when adding to a real project.

Adds support for `my_reg.sync` over a parallel comms interface for OrigenLink and OrigenSim.

Also added some simulation artifacts to the default .gitignore


priyavadan:
@ginty Model import and export export and import from a custom dir and namespace fails during Travis CI build.

ginty:
@info-rchitect, @CodyBriscoe - one thing I fixed in this PR is that test was that this test which I added a few weeks ago in response to a bug report from Cody:
https://github.com/Origen-SDK/origen/pull/244/files#diff-27ae5b59f9f8160d744e3c4d01490a3aR148

Previously it would run fine locally, but it would always fail when executed on the CI server, and I couldn't figure out why.
I found out the reason was because when the `on_flow_end` callback is called, the test program has not been closed and the final AST/model that doc helpers uses to generate the documentation has not been written yet.

The test worked fine if you had already run it before, since it picked up the last written AST/model instead, however it failed when it was the first time it was run.

In your case, what that means is that if you added a new documentation comment and then generated the program/documentation, you would not see it the first time since you would still be looking at the documentation generated via the previous version of the source code.

I doubt that explains everything you have been seeing, but possibly it might explain some of it.

The solution is to use the `program_generated` callback instead, which has been added in this PR.

PR #247, opened 2018-03-14
Hook for SMT8 Patterns

ginty:
Adds a hook to support tester drivers where a final pattern is comprised of multiple files - required for SMT8.

Also removed some test by-products which should not be checked in since they generate diffs when the tests are run by different people.


ginty:
@info-rchitect, appreciate if we can include in 0.32.

Thanks!

PR #248, opened 2018-03-19
Remove dependency on p4ruby gem

ginty:
The p4ruby gem has caused a bit of havoc within the NXP environment due to it requiring a native extension which it seems to want to download from a 3rd party location via FTP.

We could probably update our firewall rules or whatever to fix this, but it generally highlights that its not worth including non-native Ruby gems if they don't add significant value to a majority of users.

For this reason, the Perforce driver has been extracted to a plugin, making the p4ruby dependency opt-in rather than included by default.

All this means is that any applications that want to use the Perforce driver capability must now add this to their Gemfile:

~~~ruby
gem 'origen_perforce'
~~~

After that, it is business as usual.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -16,7 +16,6 @@ PATH
       log4r (~> 1.1.10, ~> 1.1)
       nanoc (~> 3.7)
       net-ldap (~> 0.13)
-      p4ruby (~> 2015.2, >= 2015.2.1313860)
       pry (~> 0.10)
       rake (~> 10)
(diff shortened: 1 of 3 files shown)


info-rchitect:
@ginty yeah I guess I really don't care how it is accessed but can we make resolving what is causing NXP to have so many issues with mixed gems (nokogiri, p4ruby) be part of v1.0 docs?  If we are successful, someone else will run into your issues.  I will take a look at it today.

PR #250, opened 2018-03-20
Dynamic Testflow Comments

CodyBriscoe:
Hijack cc method when called during program generation for creating dynamic descriptions, requires corresponding origen_testers change.

### DIFF
--- a/lib/origen/global_methods.rb
+++ b/lib/origen/global_methods.rb
@@ -6,7 +6,12 @@ module Origen
     end
 
     def c1(msg, options = {})
-      Origen.app.tester.c1(msg, options)
+      if Origen.tester.generating == :program
+        # Test Flow cc call must be on the line directly before test method
+        Origen.app.tester.c3(msg, options, caller_locations(1, 1)[0].lineno + 1)
+      else
+        Origen.app.tester.c1(msg, options)
+      end
     end
     alias_method :cc, :c1
 


info-rchitect:
@ginty do we need to add in some examples that ensure this works?

ginty:
The flow is not just merge at will if we don't get around to reviewing within a day

ginty:
This should not have been merged as is IMO.

The c3 method should be renamed to something else, the c1, c2 exiting method names denote level 1 comment, level 2 comment, etc.

info-rchitect:
@ginty so we need a way of knowing when folks will get around to stuff.  Everyone is busy so perhaps there needs to be a flag indicating a change is gating.  we would have to no problem waiting as long as it takes to get a review, then we would just move the code internal.  we have been trying to not have to do that so we stay in sync with github.  pls un-merge it and provide your review when you can get to it.  we will merge it to an internal version as it is gating.

ginty:
Hi @CodyBriscoe, thanks for your first PR and sorry that it got caught up with a discussion about how we handle code approval/merging in general!

I have reverted this, and I was going to ask you to resubmit it, however I wonder if it is really required...

As mentioned in previous comments, the c1, c2 nomenclature is meant to represent the significance of the comment, e.g. c1 comments are more important than c2, therefore they may be formatted to make them standout more in the final pattern (for example).
So, using c3 to mean 'test flow comments' does not really fit with that, and I think we should just be re-using the c1/cc method in the test flow source files.

Since the current c1 implementation hands off to OrigenTesters:

~~~ruby
Origen.app.tester.c1(msg, options)
~~~

Then I think we can handle any pattern vs. program branching over there and no changes should be required here.

What do you think?

PR #255, opened 2018-04-09
changed when Origen.current_command is available

info-rchitect:
this allows code in `boot.rb` to query `Origen.current_command`, relating to [this discussion](https://stackoverflow.com/questions/49636962/can-origen-current-command-be-nil).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -102,7 +102,7 @@ GEM
       require_all (~> 1)
     origen_updater (0.7.0)
       origen
-    parser (2.5.0.4)
+    parser (2.5.0.5)
       ast (~> 2.4.0)
     powerpack (0.1.1)
     pry (0.11.3)
--- a/lib/origen/commands.rb
+++ b/lib/origen/commands.rb
@@ -26,6 +26,9 @@ ORIGEN_COMMAND_ALIASES = {
 @command = ORIGEN_COMMAND_ALIASES[@command] || @command
 @global_commands = []
 
+# Moved here so boot.rb file can know the current command
+Origen.send :current_command=, @command
+
 # Don't log to file during the save command since we need to preserve the last log,
 # this is done as early in the process as possible so any deprecation warnings during
 # load don't trigger a new log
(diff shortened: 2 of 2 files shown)


coreyeng:
Looks fine to me. FYI, travis is failing because the developers yanked the parser version in Origen's Gemfile.lock: https://rubygems.org/gems/parser/versions/2.5.0.4

Travis won't update the gems (probably for the best) but if you do <code>bundle update parser</code> and check in the new Gemfile.lock, the Travis builds should work.

That's also why the OrigenTesters and ATP gems are failing. Same error.

info-rchitect:
thx @coreyeng . What did you mean by 'the developers yanked the parser version in Origen's Gemfile.lock'?  All I did was clone my fork, make the single edit, and commit.  I also tried `bundle update parser` and still failed `origen test`.  Any dieas?

coreyeng:
Hmm. I still don't see that a new Gemfile.lock was pushed. The problem Travis is having is external to anything you did, you were just the unlucky one how checked something in first.

When travis runs, it builds everything from scratch, so it will use whatever is in Origen's latest Gemfile.lock. However, one of the gems in the Gemfile.lock is missing, so travis can't build:

image

This is why it fails. It needs Parser version 2.5.0.4, per Origen's Gemfile.lock, but that version was yanked and can't be pulled. The travis script Stephen setup can't update missing gem versions, and instead just fails.

You should see a similar problem if you tried to rebuild your gems from scratch. Bundler won't be able to pull version 2.5.0.4 of Parser and the Gemfile.lock needs to be updated to a version Bundler can pull.

I'm not sure what 'origen test' 'is actually. Travis is just running specs. 'origen test' fails for me to, but I don't know what that is and I don't think its what Travis' issue is.

coreyeng:
Thanks @info-rchitect! Looks like that did it. Its a pretty minor thing, but I like seeing green. :smile:

info-rchitect:
@coreyeng thx for the help, did not understand that dependency.

coreyeng:
No problem. Yeah, its a weird one but its actually nice Travis starts from scratch or we wouldn't catch stuff like that as readily.

info-rchitect:
is someone besides me supposed to merge this?

coreyeng:
I think you're at liberty to merge as you see fit. I wasn't planning to merge it. If you're waiting for Ginty though, he's on vacation until next Monday.

PR #256, opened 2018-04-17
Allow self-identification in on_mode_changed callback

rlaj:
Pass self into on_mode_changed so that listeners can self-identify (and then optionally perform some task).  Mainly used when sub-models (e.g. sub-blocks) contain modes that are separate from any top-level modes (or modes of other sub-models).

### DIFF
--- a/lib/origen/model.rb
+++ b/lib/origen/model.rb
@@ -219,7 +219,7 @@ module Origen
     def current_mode=(id)
       @current_mode = id.is_a?(ChipMode) ? id.id : id
       Origen.app.listeners_for(:on_mode_changed).each do |listener|
-        listener.on_mode_changed(mode: @current_mode, klass: self)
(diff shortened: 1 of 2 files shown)


ginty:
I think klass is not the right name for this, that's normally used to represent a class whereas this is an instance. Something like model, object, instance, etc. would be better.

rlaj:
Yes, agree.  Updated to 'instance'.

ginty:
Thanks, and of course - http://origen-sdk.org/origen/guides/misc/callbacks/#on_mode_changed

rlaj:
I was updating the docs along with the source code (I can be trained;)

ginty:
Sorry! When I scanned the diff yesterday I thought the doc change was a spec change.
Indeed, you are well trained now :-)

PR #260, opened 2018-04-30
added in add_pin_groups method

info-rchitect:
add_pin_groups method accepts a block and allows a list of Symbols, Arrays, and Regexp to be used to create pin groups, with optional package(s) being passed to create the groups with package scope(s).

### DIFF
--- a/templates/web/guides/models/pins.md.erb
+++ b/templates/web/guides/models/pins.md.erb
@@ -396,6 +396,55 @@ pins(:g1).size      # => 3
 pin_groups.size     # => 1
 ~~~
 
+Pin groups can be created using a block format that allows a mix of Symbols, Arrays, and Regexp.
+
+~~~ruby
+add_pin(:ddr_bus1)
+add_pin(:ddr_bus2)
+add_pin(:ddr_bus3)
+add_pin(:ddr_clk)
+add_pin(:tdi)
+add_pin(:tdo)
+add_pin(:tms)
+add_pin(:tck)
+add_pin(:trst)
+add_pin_groups do
+  add_group :ddr_data_and_clk, /^ddr_bus/, :ddr_clk
+  add_group :jtag, [:tdo, :tdi, :tms, :tck, :trst]
+  add_group :jtag_wo_trst do
+    with :jtag
+    without :trst
+  end
+end
+
+pin_groups.ids                          # [:ddr_data_and_clk, :jtag, :jtag_wo_trst]
+pin_groups(:jtag).map(&:id)             # [:tdo, :tdi, :tms, :tck, :trst]
+pin_groups(:jtag_wo_trst).map(&:id)     # [:tdo, :tdi, :tms, :tck]      
+pin_groups(:ddr_data_and_clk).map(&:id) # [:ddr_bus1, :ddr_bus2, :ddr_bus3, :ddr_clk]     
+~~~
+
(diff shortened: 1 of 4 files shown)


templates/web/guides/models/pins.md.erb
```diff
@@ -396,6 +396,55 @@ pins(:g1).size      # => 3
...
+add_pin(:ddr_bus1)
+add_pin(:ddr_bus2)
+add_pin(:ddr_bus3)
+add_pin(:ddr_clk)
+add_pin(:tdi)
+add_pin(:tdo)
+add_pin(:tms)
+add_pin(:tck)
+add_pin(:trst)
+add_pin_groups do
```
ginty: I still don't get what the block is doing in an example like this?

A block is used like this when defining a register for example, because the outer block owns/defines the register and the inner block configures it.
Here, it seems like the outer block is doing nothing and the inner block is creating independent pin groups; subsequently the API looks weird to me.

i.e. why is it not just:

~~~ruby
add_group :ddr_data_and_clk, /^ddr_bus/, :ddr_clk
add_group :jtag, [:tdo, :tdi, :tms, :tck, :trst]
add_group :jtag_wo_trst do
  with :jtag
  without :trst
end
~~~

templates/web/guides/models/pins.md.erb
```diff
@@ -396,6 +396,55 @@ pins(:g1).size      # => 3
...
+By not passing in a package argument (as shown above), the pin group is created at the die level.  A single package can be passed as a Symbol or multipe packages can be passed as an Array, such that the pin groups are cteaed with package scope.
+
+~~~ruby
+add_pin(:uart1)
+add_pin(:uart2)
+add_pin(:uart3)
+add_pin(:rgmii1)
+add_pin(:rgmii2)
+add_pin(:rgmii3)
+add_pin_groups :bga do
```
ginty: So in this case the block wrapper makes sense since it seems to be doing something and is defining a group called :bga (though add_pin_group :bga (singular) would read better).

I assume this means the internal group definitions will create the standalone pin groups and also add them all to :bga?

ginty: Actually no, I see that there is an all_bga_pins definition which does that, so again I am confused as to what the outer block is for? 😕

info-rchitect: the outer block defines package, with default to the die level.  I think you are seeing the examples, which are very small, and not imagining a chip with 2000+ pins. for each package (could be 5+ packages per silicon revision) the block wrapper makes it very convenient to add dozens of pingroups with package scope.

ginty: Ah OK, in that case I think the API should be clearer, and in fact it should be existing `with_package` wrapper - http://origen-sdk.org/origen/guides/models/pins/#Pin_Availability_by_Package

info-rchitect: @ginty the current API does not allow multiple packages to be specified, what do you want to do?

ginty: This change should then extend that existing API rather than add a new one

info-rchitect: @ginty not trying to be argumentative but I did think I was adding to the existing API.  The add_pin_group method accepts a package as a default option and so does the add_pin_groups method.  I could change it from being an argument to a hash option if that would help.

info-rchitect:
@ginty OK well this will have to wait then as I used up my development time for this week.  if someone else wants to take it from here that would be great as well.

ginty:
Closing this due to lack of activity, obviously just re-open or create a new PR if that changes

Issue #262, opened 2018-05-03
Purpose of nil bits?


coreyeng:
I'm back to, and almost completed, spec-ing and documenting the collector. I added some miscellaneous cases to it to make it behave closer to how a hash works, e.g., setting multiple values, setting the same value multiple times (technically this is legal, but doesn't really serve any purpose and is most likely a mistake), and empty value raise exceptions.

Currently, I know the collector accepts empty values and that empty values are used, since I printed deprecation warning at first to see if they would come in when running the full specs, and the empty case is used for registers.

Looking at the specs, its used to initialize bits. Which I can see, but is there a purpose in initializing it to nil? The bit will exist, but won't Origen treat this as a 0 when actually used and wouldn't it be clearer to initialize it to 0?

Furthermore, should the collector complain about missing values? I've only seen this case and a case in RosettaStone (internal gem) that have this issue, so it doesn't seem widely used (and, for now, it won't crash, just print deprecation warnings). However, a change to the register would require a spec change.

Since its in the spec, I have no problem keeping it, but I think an improvement would be to initialize those bits to 0, then figure out whether the collector should complain about empty values or not.

For reference, this is what I mean by empty values:

~~~ruby
Origen::Registers.default_bit_meta_data do |bit|
  bit.attr_bx
  bit.attr_by = 10
  bit.attr_bz = 20
end
~~~

Currently, I would have the collector raise an error about <code>attr_bx</code> not having a value, since as a hash <code>attr_bx: </code> would not be allowed, though <code>attr_bx nil</code> would be, as its the same as <code>attr_bx: nil</code>. That's my reasoning at least.

You can see the full scope of what I plan to PR soon here: https://github.com/coreyeng/origen/blob/master/spec/utility/collector_spec.rb

info-rchitect:
@coreyeng did I miss the original thread on what this is used for?

coreyeng:
Nope, there was no original thread. The collector was something @ginty put together to use mash an options hash with an options block (allows bits to be declared using either a block or has, or both).

However, I've been putting it in stuff I've been working on since you mentioned block support for Componentable back in October of last year I think. I'm spec-ing/documenting it now so its 'official' and shouldn't change on us, potentially breaking stuff that uses it but isn't itself specs (e.g., internal gems).

coreyeng:
From the docs I'm writing:

[...]

Issue #259, opened 2018-04-25
Gems without nokogiri cannot compile web docs


info-rchitect:
Hi,

Here is the error:

~~~shell
Loading site data\u2026 
Captain! We\u2019ve been hit!

Message:

LoadError: cannot load such file -- nokogiri
Make sure the gem is added to Gemfile and run `bundle install`.

Compilation stack:

  (empty)

Stack trace:

  0. lib/bootstrap_filter.rb:1:in `require'
  1. lib/bootstrap_filter.rb:1:in `<main>'
  2. /home/bcaqueli/.origen/gems/ruby/2.3.0/gems/nanoc-3.8.0/lib/nanoc/base/source_data/code_snippet.rb:34:in `eval'
~~~

I did a recursive grep and the only difference between gems that can compile web docs and those that cannot is the lack of nokogiri being in the Gemfile.lock.

~~~text
# gem that can compile webdocs
,yserver $ grep -ri nokogiri *
Gemfile.lock:      nokogiri (>= 1.4.1)
Gemfile.lock:    nokogiri (1.8.2)
Gemfile.lock:      nokogiri (~> 1)
lbin/nokogiri:# The application 'nokogiri' is installed as part of a gem, and
lbin/nokogiri:load Gem.bin_path("nokogiri", "nokogiri")
web/lib/bootstrap_filter.rb:require 'nokogiri'
web/lib/bootstrap_filter.rb:    doc = Nokogiri::HTML.fragment(content)
web/lib/bootstrap_filter.rb:    Nokogiri::HTML(html).css("div")
web/lib/codeblocks_filter.rb:require 'nokogiri'
web/lib/codeblocks_filter.rb:    doc = Nokogiri::HTML.fragment(content)
web/lib/search_filter.rb:require 'nokogiri'
web/lib/search_filter.rb:    doc = Nokogiri::HTML(content)

# gem that cannot compile web docs
pmyserver $ grep -ri nokogiri *
lbin/nokogiri:# The application 'nokogiri' is installed as part of a gem, and
lbin/nokogiri:load Gem.bin_path("nokogiri", "nokogiri")
web/lib/bootstrap_filter.rb:require 'nokogiri'
web/lib/bootstrap_filter.rb:    doc = Nokogiri::HTML.fragment(content)
web/lib/bootstrap_filter.rb:    Nokogiri::HTML(html).css("div")
web/lib/codeblocks_filter.rb:require 'nokogiri'
[...]

PR #263, opened 2018-05-04
Spec-ing the Collector

### LINKED ISSUE #261: origen test fails is user specifies where to install gems
~~~ruby
  7) Origen.site_config Site config install directories with ENV variables set with ORIGEN_GEM_INSTALL_DIR set uses the value in ORIGEN_GEM_INSTALL_DIR instead of :gem_install_dir
     Failure/Error: expect(Origen.site_config.user_gem_dir).to eql("/env/gem/.origen/gems")
     
       expected: "/env/gem/.origen/gems"
            got: "/users/user/origen/.origen/gems"
     
       (compared using eql?)
     # ./spec/site_config_spec.rb:243:in `block (4 levels) in <top (required)>'
     # ./config/commands.rb:42:in `<top (required)>'
     # ./lib/origen/commands.rb:185:in `require'
     # ./lib/origen/commands.rb:185:in `<top (required)>'
~~~

When travis runs the same build it passes.

coreyeng:
Hi. I've been using the Collector (utility to transform a block into a hash) in more and more features, and I felt the need to properly spec and document it just so its officially part of Origen. That is most of what this PR is.

The basics of the collector are unchanged. All existing specs still pass. A few changes I made were deprecating the <code>store</code> method (it still works, but prints a deprecation warning) in favor the more Ruby-centric <code>to_h</code> and <code>to_hash</code> methods. The collector will also fail if more than one argument is given. Before, any extras would be ignored so I think its more transparent to just fail instead of dropping off parts of the value. Lastly, the collector supports auto-merging now allowing for one-liners to merge an options hash and options block. These are all documented and have accompanying specs tests. Per #262, which was just me not looking closely enough at what the register specs were doing, I've left empty method as setting <code>nil</code> value.

In addition, I added Nokogiri to the gemspec for #259 and added clearing all ENV variables along with the site config when running the site config spec, which I think will fix #261.

Thanks,

Corey


lib/origen/utility/collector.rb
```diff
@@ -1,14 +1,78 @@
...
+
+        if block_given?
+          yield self
+        end
+      end
+
+      # Legacy store method.
+      def store
+        Origen.log.deprecate 'Collector::store method was used. Please use the Ruby-centric Collector::to_h or Collector::to_hash method instead' \
+                             " Called from: #{caller[0]}"
```
ginty: I think if you call `Origen.deprecate` instead of the direct log method, it should report the caller for you like this, just FYI

coreyeng: Thanks! I'll use this next time. I didn't know the Origen module had that method as well.

PR #265, opened 2018-05-21
Remove volts as unit of measure for numeric attributes for inspect me…

info-rchitect:
…thod

some power domain attributes are current/power.


welguisz:
@info-rchitect ... Just to make sure, all you are doing with this PR is to make it work with Amps and Watts.

info-rchitect:
@welguisz yes and I am doing so by not assuming a numeric should have a 'V' appended to the end.  The column headers should provide enough context.

PR #270, opened 2018-06-20
Make regular comparison operators work

ginty:
I just found a bug in OrigenSim because I expected this to work:

~~~ruby
Origen::VersionString.new('0.10.0') < '0.8.8'
~~~

Now it does, and so do all other regular comparison operators.

The original API is still supported for masochists, but the specs have been re-written to be examples of using the improved alias methods.


ginty:
Sorry, merging this without review since it is very minor, has passing tests and I need it to fix an Origen Sim issue ASAP.

PR #271, opened 2018-06-20
Allow bit names that are similar to 'reg level attributes'

rlaj:
Any bit name that matched a reg_level_attribute ('feature', **'reset'**, 'memory', etc.) was being clobbered during register creation (using add_reg or reg do block).  Patch that adds a underscore prefix to the attributes in the class hash and handles ignoring it when creating the instance variable associated with those attributes.


rlaj:
Hmm, would swear it was passing specs yesterday.  Need to make some updates apparently...

PR #274, opened 2018-06-28
Add read/write alias methods to pin/pin collection

pderouen:
Registers are operated on using read/write.  It's common for people to attempt to use read/write with pins as well, which leads to run time errors and a search for the proper method name.  Seems intuitive to have these alias methods instead.


pderouen:
@ginty I didn't see any tests for the existing alias methods.  Is it necessary to add some tests?

pderouen:
done

pderouen:
@ginty since I pushed another update (updated the j750 workout test), I wanted to make sure you're still ok with this being merged.

ginty:
Still looks good @pderouen, thanks

Issue #108, opened 2017-07-24
Cascading Commands


coreyeng:
Currently, with the way commands are handled, its possible to overwrite commands during the plugin's import phase and the one that will be used is dependent on the load order of all the plugins. For the most part, this is okay, since there's only the Origen commands, plugin commands (with a limited number of plugins at the moment), and application commands. But, with the possible addition of PR #105 and the potential for Origen to grow larger and receive even more plugins, this may cause some headache in the future.

I've added a 'command considerations' topic in 'guides/plugins/creating.md.erb' to just discuss and warn of this behavior that I'll ship along with PR #105 when complete but longer term I'd like a better way to add and monitor commands as they are being added.

I'm alright with overriding commands and all, but I think we can handle it better than just stacking a case statement on top of itself. I'd like to switch around to using some sort of hash to keep track of the command and the plugin and warn if a global commands steps on a shared command, allow for Origen to handle the namespacing (I'll provide some shortcuts, for example nvm instead of needing c90tfs_nvm_tester), maybe give some consistency in the output of '-h' since commands can kinda do whatever there now, and still allow for an application to override a global or shared command. I hope we can head off a problem that no one has encountered yet.

I'll be sure to support the older style for backwards compatibility but I'll throw out some deprecation warnings when used.

I'll assign to this myself and work on it as I have time and, to be blunt, this will be low priority but please leave any feedback you have on the general idea or correct me if I'm misunderstanding how the commands currently work. I'm putting this here so that we can mark it as a potential pitfall and to note 
that someone (me) will work on it eventually.

coreyeng:
@ginty could you assign this to me, please??? I guess I can't assign this to myself since I don't have write access to the repo.

ginty:
Sounds great and definitely needed.

Currently we have tried to work around the issue of command clobbering by using a name space, like my_plugin:my_command, but that is very much down to individuals if they follow that or not.

I would like to see the namespacing enforced by Origen.

PR #276, opened 2018-07-02
Updates for Windows, pattern header, and failing.

coreyeng:
Hi,

I started some work on #108 but I started it on Windows. I encountered a few problems so this is PR to fix some of those. I redid the pattern headers (want this for a windows-based plugin), and added some various other features/fixes.

#### Web Compile (Windows)

Had some trouble trying to run just <code>origen web compile</code> on Windows (development only, not trying to release). There were two problems here:

First, the hostname for Windows was hardcoded to legacy FSL. Updated to (hopefully) an OS-agnostic way of finding the host. Tried on both Windows 10 (Ruby 2.5) and Linux (Ruby 2.3) and both seem to work correctly. This uses Ruby's standard <code>Socket</code> library, so should stand a good chance of working for other OSs that support Ruby (MacOS, Windows 7, other Linux distributions, etc.).

Second, some reason (still not sure exactly why…, but may related to install options?) when building the search filter during the web compilation, the encoding was getting messed up. This was leading to search filter JSON that resembled an executable more so than text and was blowing up the file size to 800 MB+ (yep, 800 MB) after only running 15-ish templates and Ruby was giving up (returning errors about files too large). Smaller websites would work, but larger ones (like origen_core) stood no chance. Forced the encoding in the nanoc config to read in UTF-8 (the guides [here](https://nanoc.ws/doc/troubleshooting/#character-encoding-issues) show it forces UTF-8 on the output) to no avail, but left it in for clarity and to make sure the input matches the output. However, was able to fix by forcing the <code>file.read</code> in the search filter to force reading with UTF-8 encoding. Resulting size is much better and matches its Linux counterpart. I've not noticed any adverse effects, but included details here in case someone notices something is awry.

#### Origen-Core Specs Robustness

Noticed a semi-dangerous rspec issue that if there's local modifications and the production target is set (default), the specs will pass without them all being run. I say semi-dangerous, because Travis will still catch this (since everything will be checked in when it runs, so this problem won't be seen), but it means that during development we'll see 'pass' even though that may not actually be the case. So, really just a development issue. For example:

image

[Based on this issue](https://github.com/rspec/rspec-core/issues/2246), this is the behavior of rspec, so there's no workarounds in rspec itself. This would need to be addressed in the test case. However, it means anything that checks the production target, whether knowingly or not, will need to build in a SystemExit-rescue branch or change the mode to debug to avoid bailing on the remaining specs.

To get around this, I made a few changes to the spec command:
1. Origen.mode is set to debug for all specs runs. It is changed back to whatever it was before after the specs tests are run.
2. The specs command will also catch all <code>SystemExit</code> exceptions and print a message saying 'unexpected system exit found' and fail with an exit status 1. <code>SystemExit</code> can still be used in code, but they must be caught by the test case itself, or else it will fail rspec.

Just for clarity, I'm not trying to disallow <code>exit</code> from being used, just stop it from prematurely killing the specs process.

#### Origen-Core Specs Command

Also added some flexibility to the spec command. Now, it can accept arguments as a substring of filenames to match. For example, if I only want to run the file site_config_spec, I can do <code>origen specs site_config</code>. Since there's only one file in <code>spec/</code> path at this time that matches, only that file will be run.

Note, it can accept arguments as either the path to a spec file or part of a spec file name to run. No arguments remains the same (runs full suite). E.g.: Origen spec site_config will match spec/site_config.rb and run specs with only that file. However, <code>origen specs version</code> will run both files <code>spec/version_history_spec</code> and <code>version_spec</code>.

Please give me feedback for the direction I'm taking the specs command. I'd like to make <code>specs</code> and <code>examples</code> core commands in future Origen versions (the application will still be able to override if it needs to though). I've grown to really appreciate regression testing (even though it takes a bunch of time to do thoroughly) and have been saved by it more than once so I hope to standardize its usage, conventions, etc. across Origen plugins.

Ironically, I didn't do any regression testing for this, and I didn't update the guides since its <code>origen_core</code> specific at the moment, but I did fill out a <code>origen specs -h</code> command that goes over this.

### Failing

I've had a problem or two with plugins using <code>exit</code> without giving much details as to what the problem is. Sometimes it just reads 'unexpected error!', and has several branches that exit with that same message, so grepping or searching the source gives five or six results and a development environment is needed to figure out which path its taking. I've definitely been guilty of this myself, so I added an <code>Origen.app.fail</code> method that will print which app/plugin is responsible for the fail. A <code>message</code> and <code>exception_class</code> option can be given to customize the exception raised. <code>Origen.app.fail!</code> will use <code>exit</code> to cleanly exit without a stack trace, <b>unless</b> <code>-d</code> (<code>--debug</code>) is used. Then, it will raise an exception, showing the same message and the stack trace. I think this will help debug if less-than-ideal error messages occur or can help end users give better debug reports. This feature is documented and spec-ed.

#### Miscellaneous

* Added YARDOC for the [Collector](http://origen-sdk.org/origen/guides/misc/utilities/#Collector). Meant to do this when I originally updated it... but forgot to.
* Default target in origen_core isn't populating on Windows as a symlink (at least not without being admin and using some options). It just had <code>debug.rb</code>, which loading doesn't work. You get the <code>NoMethodError debug</code> when loading. Changed this to a regular file and loads the production target. This is only applicable for origen_core. Doesn't affect other apps. No change to how the default target works at the Origen level.
* The site config spec was failing on Windows for two reasons. First: the <code>/proj</code> was expanding to <code>C:\proj</code>, and second: trying to expand the forward slash actually generates a path. Added in some fixes for both of these. Confirmed site config spec passes on both Windows and Linux with these changes.
* DesignSync 2017 changed its naming convention (sigh, …again). Updated for this. Fixes the issue of the DesignSync version string being registered as a changed file, even though it doesn't exist as a file at all.
* Added #running_interactively? to Origen. E.g.: <code>Origen.running_interactively? #=> true/false</code>


coreyeng:
Two notes I forgot to mention:

1. I ran this on Ruby 2.5 (Windows 10), so just a bit a look ahead for when we require a later Ruby version.
2. There's a bug that has something to do with exiting using <code>Origen.app.fail!</code>, but only when running interactively. There's an <code>uncaught throw :IRB_EXIT (UncaughtThrowError)</code> being raised that I'm not sure about and some preliminary web searches didn't yield any help. I plan to open this as an issue if this PR is accepted, but didn't want to hold it up since I'm not sure where to start tbh. Also, the <i>spirit</i> of the method is preserved, and it even exits interactive mode! It also isn't an issue for non-interactive mode runs. I think a workaround would be to add a case when running interactively and use <code>exit!</code> instead, if no alternative is found or if we get complaints about the error.

Here's the error, if anyone has any recommendations. Happens on both Windows (2.5) and Linux (2.3), so not OS or Ruby version dependent it would seem:

image

ginty:
Sorry @coreyeng, looks like this fell off the radar, thanks for bumping (don't much like Github's recent-ish changes to the org-overview page, seems like it only shows you an overview of things you have touched recently instead of a truly org-level overview).

Anyway, this all LGTM.

I agree with the direction you are going with the regression command and agree that these should be further integrated into Origen to reduce the amount of app-side logic in future.

On the failing topic, I think that any code which calls exit should also be responsible for outputting a useful error message to the console, so we should look out for that in future peer review. No issues, with the additional APIs you are adding to help with this though.

Overall a nice set of updates, thanks!

coreyeng:
Thanks for the quick feedback!

coreyeng:
Hi @ginty. Well, in kind of an embarrassing moment, I completely left out the changes to pattern headers. Looking back in my notes I made for this PR, I omitted it completely... Anyway, I changed the pattern header and now allow applications, plugins, and the current plugin to inject pattern header comments. The current method <code>config.pattern_header do ...</code> still works, but it prints a warning now. I also print the current mode into the pattern header.

[I even wrote up docs for this.](https://github.com/Origen-SDK/origen/pull/276/commits/6da6f35e038c4024318baf17b269296b79a1d0f1#diff-31146ec0e14ef0d7a483ab0376378227) Very sorry, but would you mind taking a quick look at that as well? Thanks!

ginty:
Hi @coreyeng, yes I noticed that yesterday when looking through the code and I read through your doc for it, looks good too.

PR #282, opened 2018-08-14
Logger improvements

ginty:
This PR does the following:

- Re-implements the logger using Ruby's built in Ruby::Logger rather than relying on a 3rd party gem which doesn't really add anything (log4r)
- Adds better logger tests which actually test the output rather than an internal interception. @coreyeng I had to rework some of your tests in the same way.
- Removed the rolling log file, I don't think anyone really used this and it just added un-necessary complexity. One previous revision of log/last.txt is maintained for that occasional time where you wish you hadn't just overwritten it by running Origen again.
- Improved documentation of the logger, including the new features...
- A log file will now be generated per-job, the real benefactor of this should be OrigenSim, and you will now get a permanent log file for each simulation you run:

Each test pattern, or program flow which is generated is an example of a job, as is each file that is compiled by Origen.
 The console output will inform you of the location of the job specific files that are generated, as shown
in this example:
 ~~~text
[INFO]       0.005[0.005]    || **********************************************************************
[INFO]       0.031[0.026]    || Generating...  /home/stephen/Code/github/origen/spec/patterns/atp/nvm_j750_workout.atp
[INFO]       0.506[0.474]    ||
[INFO]       0.509[0.003]    || Pattern vectors: 146
[INFO]       0.511[0.003]    || Execution time : 0.019288
[INFO]       0.512[0.001]    || ----------------------------------------------------------------------
[INFO]       0.576[0.064]    || Log file written to: /home/stephen/Code/github/origen/log/pattern_generator/debug/j750_workout.txt
[INFO]       0.584[0.008]    || **********************************************************************
[SUCCESS]    0.584[0.001]    || Total patterns:   1
[SUCCESS]    0.585[0.001]    || Total vectors:    146
[SUCCESS]    0.586[0.001]    || Total duration:   0.019288
[SUCCESS]    0.587[0.001]    || New patterns:     0
[SUCCESS]    0.588[0.001]    || Changed patterns: 0
[INFO]       0.589[0.001]    ||
[INFO]       0.589[0.001]    || **********************************************************************
~~~
 The location of the job-specific log files is: `log/<job type>/<target>/<filename>.txt`.

 No old copy of job-specific log files is automatically maintained and these will be overwritten the
next time Origen is asked to perform the same job.

- Custom log files are also now supported:

Custom log files can be used to collect a specific type of data in one place.
 Calling `Origen.log.<my_log_name>` in place of one of the standard message types will cause the
given message to be logged to a custom file instead.
 For example, these messages would be written to `log/accesses.txt`:
 ~~~ruby
Origen.log.accesses "An access was made with this value: #{my_val}"
~~~
 By default, such messages will be formatted as per standard Origen log messages, with the message type
and timestamp appended to the front.
 To output the message without this additional formatting being applied, add the `format: false` option:
 ~~~ruby
Origen.log.accesses "An access was made with this value: #{my_val}", format: false
~~~
 Custom log messages are never output to the console and will be captured in the applicable log file only.
 One old copy of the file will be maintained in the same way as it is done for `log/last.txt`.


pderouen:
I'm in favor of whittling away at 3rd party gem dependencies.  :+1:

PR #283, opened 2018-08-16
Feature/lsf options

priyavadan:
Initial commit for this that I would like merged in. 

Our IT has been hounding me for using too much resources for my sims, so needed a way to pass in the number of cores requested.

I will probably enhance it to include a request for the amount of ram and swap space you can request as well.

### DIFF
--- a/lib/origen/application/lsf.rb
+++ b/lib/origen/application/lsf.rb
@@ -38,8 +38,8 @@ module Origen
         def initialize
           @group = nil
           @project = 'msg.te'
-          @resource = 'rhel6' # Defaults to new NxDI LSF strings
-          @queue = 'shortq'   # Defaults to new NxDI LSF Strings
+          @resource = 'linux'
+          @queue = 'short'
           @debug = false
           @cores = '1'
         end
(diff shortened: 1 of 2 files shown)


lib/origen/application/lsf.rb
```diff
@@ -18,13 +18,30 @@ class Configuration
...
+        #         distributed	          normal		Run jobs than span multiple hosts.
+        #         -			  prio	                High prio queue with low slot count, useful if you don't have slots available in normal queue. See PrioritizingMyJobs.
+        #         -			  ondemand     		On-Demand Servers to satisfy urgent and short-term (2 weeks or less) customer compute requirements.
+        #         -			  wam	  		WAM cron processing
+        #         -			  grid	     		Low-priority batch jobs (random sim, regressions, etc). Access to all spare CPU cycles.
+        attr_accessor :cores
 
         def initialize
           @group = nil
           @project = 'msg.te'
```
ginty: Think we should change the default project to 'origen' while we are at it

priyavadan: Requested origen project string, waiting on approval.

priyavadan:
updated so that LSF uses site config variables. Will start another PR next week to add more features, will merge this for now.

PR #285, opened 2018-08-22
Fix Git remote population

ginty:
Population of remotes via Git seemed to be broken when running with older versions of Git (v1).

This patch does the following:

- Comparison of Git version identifiers like 'HEAD' and 'master' would cause an error when they were wrapped in an `Origen::VersionString`. This is likely a regression introduced a while ago. A test for it has been added and a patch to make the `==` comparison fall back to a regular string comparison when the value is not a compliant semantic version string.

- I don't think `git 'remote remove origin', verbose: false, check_errors: false` really works with `v1` Git, so I was getting `origin already exists` type of errors when trying to add the origin remote during an initial remote population. This patch will ignore such errors, which I think is probably preferable in most cases.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,4 @@
-source 'http://rubygems.org'
+source 'https://rubygems.org'
 
 # Development dependencies
 gem 'coveralls', require: false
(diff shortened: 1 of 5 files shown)


ginty:
Yep, good point @arkohler, now added

PR #286, opened 2018-08-23
Added numeric wait helpers

ginty:
Added helpers to create pattern wait durations like this:

~~~ruby
10.ms!    # Shorthand for tester.wait(time_in_ms: 10)
~~~

The driver for adding this is to make it quick and easy to advance time while in an OrigenSim interactive session.


rlaj:
looks good.  i guess these would never be called if Origen.app.tester = nil

ginty:
@rlaj, should be OK I think, it is implemented the same as `10.cycles` which also has no explicit check for such a cse

Issue #288, opened 2018-08-30
Bootsnap gem for app loading performance


info-rchitect:
https://github.com/Shopify/bootsnap

any thoughts?

pderouen:
This comment (first sentence in the section) seems to imply that it doesn't work on Windows which would be a show stopper:
https://github.com/Shopify/bootsnap#usage

info-rchitect:
@pderouen Why would that be a show stopper?  I think Bootsnap is application specific.

ginty:
This does actually seem to work, seems to speed up the time to load the app and target in one of my apps by ~2.2s, going from 5.5s down to 3.2s:

~~~text
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m5.427s
user    0m1.250s
sys     0m3.891s
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m5.463s
user    0m1.297s
sys     0m3.906s
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m5.613s
user    0m1.422s
sys     0m3.828s
~~~

Now with bootsnap:

~~~text
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m3.160s
user    0m0.859s
sys     0m2.109s
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m3.232s
user    0m0.766s
sys     0m2.203s
stephen@NXL78024:~/Code/bitbucket/c402t_nvm_tester$ time origen nvm:build
App and target loaded!

real    0m3.404s
user    0m0.906s
sys     0m2.172s
~~~

info-rchitect:
@ginty thx for trying it, I was going to get to it in the next week or two during a refactor.  Are you going to use it?

ginty:
I think I would vote to add it to Origen based on those performance numbers.

It really needs to be added to Origen though and not at app-level so that you get the benefit of it when loading Origen and your app's plugins.

It's easy enough to inhibit it from kicking in on Windows, the only thing is that we need to make sure that including bootsnap as an Origen dependency does not break it on Windows.

@pderouen, could you please check if running 'gem install bootsnap' works on Windows or not?

[...]

PR #293, opened 2018-09-24
Add bootsnap

ginty:
Adds the bootsnap gem to Origen to speed up application boot times - https://github.com/Shopify/bootsnap

This is on by default, but can be disabled at site/app/user level via a new site config parameter - `use_bootsnap`.
It is not enabled on Windows due to the bootsnap gem not supporting it.

From experiments in #288, this is expected to speed up Origen application boot times by several seconds and should be a noticeable speed improvement to the user.

Note that it is slower the first time it is run while the cache is created in `tmp/cache`.


info-rchitect:
@ginty seems too simple for what is gained but just shows there are clever folks in the open space universe.  approved

pderouen:
@ginty I just ran from a Windows 7 install with bootsnap enabled. "origen test" passed.

pderouen:
Here's the times I measured executing "origen -v" inside an app (using powershell measure-command)

Without bootsnap
TotalSeconds      : 12.7561151

With bootsnap
TotalSeconds      : 6.1414542

Back to without bootsnap
TotalSeconds      : 12.6434103

ginty:
@pderouen, I guess bootsnap helps a bit, that's pretty bad numbers though :-(

pderouen:
It was actually a little slower with bootsnap enabled.  But, it loaded and ran just fine and I could see the cache it created. Should we enable it on Windows?  It wouldn't actually be used unless site config or the app enables it, right?  We'd need to update the windows install to include the bootsnap gem install.

ginty:
@info-rchitect, does Pry work OK with this update? I saw some comments in the bootsnap issues which suggest that it may not.

info-rchitect:
@ginty I added bootsnap to my app and have been able to use pry (step, next, etc.) alright.

ginty:
OK, thanks @info-rchitect 

Anyone object to going ahead and releasing this?

chrisnappi:
No, go for it.

pderouen:
I have no issue with it. Do any of the docs need to be updated?

ginty:
> Do any of the docs need to be updated?

Don't think so, this _should_ be a low level implementation detail that users never need to worry about.

If we find that this is not the case and there are some users who need to turn it off, we can reconsider the default setting and how we want to document it.

PR #294, opened 2018-09-26
skip reset_val export if :undefined

pderouen:


ginty:
Should it not generate the following in this case?

~~~ruby
reset: :undefined
~~~

I think that's how you would arrive in this state in the first place unless I'm mistaken, because an un-declared reset value defaults to 0 I think.

ginty:
In fact, I would implement it as any symbol value being written out as there may be others - I think `:memory` might be allowed for example.

~~~ruby
if bits.reset_val.is_a?(Symbol)
  line << ", reset: :#{bits.reset_val}"
else
  line << ", reset: #{bits.reset_val.to_hex}"
end
~~~

PR #296, opened 2018-10-22
Site config enhancements

coreyeng:
Hello,

This PR includes a few changes to expand Origen's site customization capability.

##### Site Config Refactor

Refactored the site config to follow more object-oriented principles. Makes it a bit easier to follow I think and should be easier to add more advanced functionality if needed.

I didn't update any specs cases for this, but I did need to correct one of the helpers. I originally meta-programmed a helper for test purposes but never corrected it to use the actual API method. Corrected this helper and all specs pass.

##### Centralized Site Config

Added support for a centralized site config. This site config is a http address that should be accessible by any networked machine, circumventing any issues with the site config paths not being reachable (e.g., running on a Windows PC when all the site configs are located in Linux, or having to deal with multiple networks not sharing/mirroring the same Origen installation directories).

To knock down any network slowdowns, or to facilitate running offline, the site configs should be cached. The cache directory can be set by <code>centralized_site_config_cache_dir</code>, and defaults to <code>~/.origen/cache/site_config</code>. The refresh time can be set with <code>centralized_site_config_refresh</code>. This setting is in hours but accepts decimal values and defaults to 24.

A bit of a paradox can occur though if the centralized site config sets the cache location. This ends up with the thing to be cached trying to set its cache location and actually results in the site config never seeing the cached config until it's too late. To circumvent this, any centralized site config will ignore attempts to set the site config cache. An error will printed (it won't fail though) and the value will be removed and ignored. Since refreshes take place at the end of the boot process, this doesn't have the same issue, so the centralized site config can set its own refresh settings.

Note that the cache location is not tied to home_dir in the way gem_install_dir or user_install_dir are. This would end up having the same problem since the centralized config is allowed to set those parameters but the default site config cache root and default home_dir coincide to the same directory.

Once the centralized site config is read, it will be injected towards the beginning of the load chain. More specifically, it will be the placed right after the defaults from Origen, meaning that it will be the lowest priority of any user-set configs.

Docs were updated to include the centralized site config and how to use it. I didn't add any specs for it though. Not sure if this is an acceptable waiver from specs or not. It would be possible to setup a site config in origen_core_support and access it directly from github though if we feel specs are needed.

##### Site Config ERB Support

Added support for the site config to support ERB, allowing for a single site config to be parsed dynamically. The same configuration ('trim mode') used by the compiler is used here ('%<>'). Since I like explicitness when possible, and to keep with the convention of using .erb, I set up ERB to only be active if the site config ends with .erb. That is, origen_site_config.yml will NOT be parsed as ERB, but origen_site_config.yml.erb will be. I also added sample one to the core.

The exception to this is the centralized site config. Since this could be some arbitrary web address, there's no naming convention and it will be parsed as an ERB.

##### 'origen site' Command

Added a site command (<code>origen site \<code\></code>) to help de-mystify what the site config is doing. For the most part, this will only be needed by site administrators, and can be useful when debugging site or boot issues.

The site command is documented, but here's a quick summary:

* <code>env</code>: Prints out the current site config variables and their values. Based off of Linux's <code>env</code> command. Use <code>env [vars]</code> to list only those variables. In Linux, this is similar to just piping the output to grep, but since that isn't easily doable in windows' CMD or powershell, it's given here as an optional use case. The [vars] are also parsed as regexes, so 'lsf' will return everything matching '*lsf*' and 'origen site env lsf ldap' will show all lsf and ldap vars.
* <code>configs</code>: Prints out a listing of the current config files found from highest to lowest priority, and with index numbers (for inspect_config)
* <code>inspect_config <config> </code>: Inspects and prints a single config. This will show all values in this config, any active (highest-level) values from the this config, as well some other details. Input to this is an index number.
* <code>inspect_variable <var></code>: Prints the tracing of a single variable through all the site configs.
* <code>refresh</code>: Forces a refresh of the centralized site config.

The only one of these that's really meant for end users is <code>refresh</code>, but since I was adding this I figured I might as well add the others since I've found them useful over my time working on site issues.

Just a note though, this environment shows only the site config settings. This may cause some confusion for things like the LSF which can have a site config variable, but the application or plugin can override, meaning that <code>origen site inspect_variable lsf_project</code> may not give you the value actually used by the LSF. 

##### Mailer Authentication

A few changes were thrown into the mailer as well. Within our company, we found a more generalized SMTP server exists, but it requires authentication to use so I added some options to allow for some basic authentication to work. I filled in the mailer docs on the utilities page to cover the setup.


origen_site_config.yml
```diff
@@ -12,6 +12,21 @@
 #  # Or a reference to a plugin's Git repository can be used
 #  - "http://bitbucket.mycompany.net/origen/my_app_generators.git"
 
+# Centralized Site Config
+# Origen 0.36+ supports a centralized site cnnfig. This site cnofig will be implicitly read and injected into the site
```
ginty: Had to pull you up on something :-)
Typo, cnnfig.

coreyeng:
Thanks @ginty! yeah, I seem to always let at least one typo slip through... though that comment needed some rephrasing anyway so I'm really glad you caught it!

PR #301, opened 2018-11-12
Revert including the entire Origen module in the site config. Created…

coreyeng:
Reverted including the entire Origen module in the site config as this was causing some problems in a feature @ginty is working on. Created a <code>boot_api.rb</code> file to house helpers that site config ERB templates can use instead. Its very small but can grow as/if needed.

Updated the docs to cover ERB templating in the site config, since this slipped through the original PR (was only discussed with regards to a centralized config)

### DIFF
--- a/templates/web/guides/starting/company.md.erb
+++ b/templates/web/guides/starting/company.md.erb
@@ -45,6 +45,24 @@ and edit it as required.
 The values present in this default file are the ones that will be applied by default in
 an installation where no custom configs are present.
 
+#### Dynamic Configuration Files
+
+Origen supports a static <code>origen_site_config.yml</code>, but also supports a template <code>origen_site_config.yml.erb</code>.
+If Origen encounters a site config with the extension <code>.yml.erb</code>, it will invoke the [ERB](https://ruby-doc.org/stdlib-2.5.3/libdoc/erb/rdoc/ERB.html)
+template, adding some dynamic nature to the configuration. With this, the same configuration file can be used throughout
+but can set the site according to the current operating system, the hostname, or anything else. If you are not familiar with
+ERB templating, you can view a [quickstart guide on the Origen docs](https://origen-sdk.org/origen/guides/compiler/creating/).
+
+The site config will automatically include Origen's [Boot API](https://github.com/Origen-SDK/origen/blob/master/lib/origen/boot_api.rb),
+and can be used freely out of the gate. Other methods and libraries can be added as needed to the ERB itself (or to helper
+<code>.rb</code> files then required by the ERB).
+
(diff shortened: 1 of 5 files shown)


templates/web/guides/starting/company.md.erb
```diff
@@ -45,6 +45,24 @@ and edit it as required.
...
 
+#### Dynamic Configuration Files
+
+Origen supports a static <code>origen_site_config.yml</code>, but also supports a template <code>origen_site_config.yml.erb</code>.
+If Origen encounters a site config with the extension <code>.yml.erb</code>, it will invoke the [ERB](https://ruby-doc.org/stdlib-2.5.3/libdoc/erb/rdoc/ERB.html)
+template, adding some dynamic nature to the configuration. With this, the same configuration file can be used throughout
+but can set the site according to the current operating system, the hostname, or anything else. If you are not familiar with
+ERB templating, you can view a [quickstart guide on the Origen docs](https://origen-sdk.org/origen/guides/compiler/creating/).
+
+The site config will automatically include Origen's [Boot API](https://github.com/Origen-SDK/origen/blob/master/lib/origen/boot_api.rb),
```
ginty: > Other methods and libraries...

I really don't like this comment, the site_config is a parameters file to configure Origen for a given host environment, buts that's all it is.
In the course of using it, we have recognized that even within the walls of a company there can be variations in the environment, and so we've added the ability to make the parameters dynamic based on operating system, host, etc. to allow us to have a single config that works everywhere.
However, that's as far as it goes and there should be no suggestion of anyone including or adding any elaborate code to the site configs.
At best, including a large library like Origen will simply slow boot up times since it will end up getting required twice, at worst it could cause failure to boot since you may require something here and then later find that Bundler complains because the runtime environment depends on version X of a lib but we have already required version Y.

So in fact, I would rather we add a warning box that says something like:

    You should not require any gems (including Origen) within your ERB code. The amount of logic should be kept 
    to a minimum and only be used to customize parameter values to the environment OS or a particular host.

Should we also mention here that the `Origen.os` helpers are already included and available to be used in site config templates?

ginty: I think `should not require any gems` is the right instruction. `socket` is not a gem is it, its part of Ruby stdlib (I think?), which is fine to require - but would the average engineer know the difference?
So I would rather be on the side of being overly discouraging here than overly encouraging.

coreyeng: Oh, it is part of standard lib. I never really checked where it was before.

~~~
but would the average engineer know the difference?
~~~

You'd think... but I just messed that up. :p

PR #302, opened 2018-11-13
Pin group offsets

coreyeng:
I didn't see any existing option to model pin groups whose indexes don't start at 0, so added an <code>:offset</code> option to <code>#add_pins</code> for this.

This was actually noticed from using OrigenSim where my <code>origen.v</code> file wasn't correctly mapping <code>ptc[19:1]</code> and I was having to manually adjust all the pin drivers. Nothing a find & replace couldn't fix, but still seemed like something to fix upstream. Traced this back to <code>#add_pins</code> not having an option for an offset.

In general, we don't see this very much here and up until the introduction of OrigenSim I never added entire pin groups before anyway, so this just surfaced.

If this is accepted, I'll update OrigenVerilog to pass this option to <code>#add_pins</code>.

### DIFF
--- a/lib/origen/pins.rb
+++ b/lib/origen/pins.rb
@@ -229,21 +229,23 @@ module Origen
 
         rtl_name = options[:rtl_name]
         force = options[:force]
+        offset = options.delete(:offset) || 0
         options.delete(:size).times do |i|
-          options[:name] = "#{id}#{i}".to_sym
-          options[:rtl_name] = "#{rtl_name}#{i}".to_sym if rtl_name
-          options[:force] = force[i] if force
+          pin_index = offset + i
+          options[:name] = "#{id}#{pin_index}".to_sym
+          options[:rtl_name] = "#{rtl_name}#{pin_index}".to_sym if rtl_name
+          options[:force] = force[pin_index] if force
 
           if power_pin
-            group[i] = PowerPin.new(i, self, options)
+            group[i] = PowerPin.new(pin_index, self, options)
           elsif ground_pin
-            group[i] = GroundPin.new(i, self, options)
+            group[i] = GroundPin.new(pin_index, self, options)
           elsif virtual_pin
-            group[i] = VirtualPin.new(i, self, options)
+            group[i] = VirtualPin.new(pin_index, self, options)
           elsif other_pin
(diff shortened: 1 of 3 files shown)


redxeth:
Looks good to me.

ginty:
Do such pin groups work as they should?

e.g. do tests like this work and/or how should they work? - https://github.com/Origen-SDK/origen/blob/master/spec/pins_v3_spec.rb#L431

coreyeng:
Hi @ginty, it works the way I think it should based on array indexing. For example:

~~~ruby
dut.add_pins(:pta, size: 4, offset: 1)
dut.pins.keys.select { |p| p.to_s.include?('pta') }
  #=> [:pta4, :pta3, :pta2, :pta1]

dut.pins(:pta)[0].name
  #=> :pta1
dut.pins(:pta)[4]
  #=> nil

dut.pins(:pta).data
  #=> 0
dutpins(:pta)[0].write(1)
  #=> 1
dut.pins(:pta).data
  #=> 1
dutpins(:pta)[2].write(1)
  #=> 1
dut.pins(:pta).data
  #=> 5

dut.pins(:pta)[0, 1, 2, 3].write(0xA)
  #=> (pin collection)
dut.pins(:pta).data.to_hex
  #=> "0xA"

dut.pins(:pta).write(0x5)
  #=> (pin collection)
dut.pins(:pta).data.to_hex
  #=> "0x5"
~~~

To me, that makes sense since <code>dut.pins(:ptc)[i]</code>mimics array indexing (even though its technically a pin collection), and the first element of that is the first pin in the group, or <code>ptc1</code>. I could also see that being used as a means to get the first pin, regardless of the offset.

Is that what you were expecting, or were you expecting it to be mapped to the index number? (e.g. <code>dut.pins(:pta)[0] #=> nil; dut.pins(:pta)[1].name #=> :pta1</code>)

ginty:
That's OK with me, just checking you have considered this.

coreyeng:
Thanks! Fwiw, we could always manually add pins, say, <code>ptc1-19</code> manually, then throw them all in a pin group <code>ptc</code> to get this same effect. I've had that in an application for 2-ish months and its been fine. This is just allowing Origen to do it for us at add pin definition time.

PR #304, opened 2018-11-14
Reg access fixes

chrishume:
Expanded Origen::Registers::Bit ACCESS_CODES to cover access attributes needed by CrossOrigen.  Added attr_readers for :base_access, :read_action, and :mod_write_value.  Expanded Bit access spec testing and added a placeholder for read/write spec testing (requires CoreSupport update before enabling).  Corrected :readable and :writable attributes for :rc, :rs, :rowz, :wos, :woc, and :wo1 access types.

On the table for future update:  add spec testing that readability and write-ability is correctly implemented for each access type.


ginty:
Hi @chrishume, the implementation here looks really nice, thanks for adding it.

However, I've merged in your update to origen_core_support and I've enabled the ACCESS_CODES test that you added here, however it is failing.

Could you take a look and update it as required?

As soon as you have done that I will merge and release this, thanks!

chrishume:
Hi @ginty, I've updated the expectations for the ACCESS_CODES test.  As the comments note, the behavior of the bits in Origen is actually incorrect.  I didn't see any understanding within the Bit model that the assorted "clear" mechanisms would work, other than :w1c (similarly the "set" mechanisms don't look to be fully supported).  They'll need to get fixed in a later update.

I'm going offline for several days, so if a change needs to be made to :w1c or any of the other access types, please feel free to make it.  I can begin the debug / fix process on Nov. 27th.

ginty:
Thanks for the update @chrishume.

Yes for sure only the primary access types (e.g. :rw and :ro) have their behavior modeled correctly. No need to couple adding behavior for the other ones with this change though.

info-rchitect:
@chrishume I was wondering when more advanced ACCESS_CODES would need to addressed 👍

PR #305, opened 2018-11-15
Add application archive and sandbox mode

ginty:
This update adds the following:

- Overhaul of how Origen boots within an application (necessary to support the following features)
- Sandbox mode, where an application can be configured to install all necessary gems to itself (including Git references), guaranteeing it will boot out-the-box for casual users of the app but at the expense of larger repo size
- Archive command (application-level command), creates a versioned tar ball of an app which contains everything it needs, allowing it to be extracted and used without any external gem dependencies, i.e. doesn't require network access to run.
- Extract command (global command), extracts an application archive created by above

The new boot system is significantly better than the original and should by itself be much more stable for users, eliminating the need for the fix_my_workspace script to recover un-bootable application states.
Adding sandbox mode on-top should result in bullet-proof distributions of apps to user communities.
The archive feature adds the ability to either preserve key snapshots of your apps, or to allow it to be used within a locked down environment which doesn't have network access.

Still to be added:

- Documentation, I will transfer the key points from this description into the guides before merging
- The prompts mention an origen command called 'origen new binstub', but this has not been added yet. This will be done in the app_dir branch which has the (significant) underlying infrastructure required to enable it.
- The app generators templates will need to be added to make them come setup for this new system out the box, however I propose to postpone that activity since they are going to require a major overhaul anyone once the app dir update is available.

### How To Use

When your application is updated with this Origen version, it will begin prompting you to upgrade it to the new boot system, follow the instructions to upgrade, which are simply to run this command:

~~~
origen setup
~~~

Once upgraded to the new system, **the ./lbin directory should now be checked into your app**.

You may then start to see prompts to remove fix_my_workspace and you should do this.

You can then use your app as normal. The main change you might notice is that you should no longer ever see a prompt telling you to bundle install, that should be automated now (which was intended in the original boot system, but which didn't work in practice).

If you want to create an archive of your app run:

~~~
origen archive
~~~

To extract such an archive you would run the following command outside of any app workspace:

~~~
origen extract path/to/my_app_archive.origen
~~~

Since this is a global command you would need to setup your user global install to point to this Git branch if you want to test that out, but rest assured it works.

To setup your app to install gems to itself, i.e. to configure it for sandbox mode:

~~~
origen archive --local
~~~

After that just use as normal but all gem installations will be in `./vendor/gems` instead of your central gem store, usually `~/.origen/gems`.
It is expected that in this configuration you would add `./vendor/gems` to your revision control system and check it in.

A new user could then check out your app and find that the `origen` command works out-the-box with no gem installations required.

### New workspace setup

The same recommendation still exists as before, that is to always run an Origen command (any) after creating a new workspace before you ever invoke the `bundle` command from the command line.

The new system is somewhat improved in that any command you launch initially will go on to run after installing gems (if required), whereas previously it could exit early.

So, the general recommendation for dealing with a fresh workspace is to just begin using Origen and stay away from 'bundle'.

I hope that with the new features added here manual invocation of `bundle install` will be completely removed the minds of the average users.

### Replacement for fix_my_workspace 

That script was always a band aid over underlying weakness in Origen's boot system, and basically should no longer be required.

The `origen setup` command can be safely run again at any time and it could be thought of as a replacement for fix_my_workspace.

The only way I know can think of to end up a locked out state could be if the user manually runs bundle install with the `--binstubs` option, so don't do that.

In that case it could be resolved by checking out your `lbin` directory, which should now be in version control and would could be overwritten by the above bundle operation.

### Some Implementation Details

The fundamental change to the boot system is that Bundler is now invoked by Origen, rather than Origen sitting behind Bundler.

When launching Origen in the past via `./lbin/origen` you had to have a valid local gem environment before Origen would be called. This meant that Origen couldn't help in cases where the gem environment was invalid in some way and an external script (i.e. `fix_my_workspace`) had to step in.
So the new system uses an Origen-owned binstub (`./lbin/origen`) rather than a Bundler-owned binstub.

In the course of this update I took an in-depth look at how rails is booted and it takes a similar approach. Overall the rails system is simpler than what we have here, but that is due to rails targeting a user community who are happy to understand the details about bundler and use it manually, whereas we are targeting a user community who wish to be oblivious to things like that. 

The main Origen executable has been moved from `bin/origen` to `lib/origen/boot.rb`, this is to give the binstub a way to tell what boot system it is dealing with by checking if the new file exists.
The code for booting outside of an app has been preserved as-is (aside from some minor clean-up refactoring), but it has been completely de-coupled from the code for booting inside of an app. This was done to avoid breaking the global execution space when changing the in-app space.
It is possible that some of these changes could be applied to global execution too, however that is outside the scope of this update and generally I don't care about that so much because it is the domain of power users and what we have today works fine.

Perhaps a slightly controversial part of this implementation is that it takes control of Bundler's configuration file (`.bundle/config` within the app) and writes it directly.
When developing this I found multiple times when Bundler seemed to be flat out ignoring configuration variables put into `ENV` that are supposed to override what is in `./bundle/config`.
Taking control of the config file resulted in more robust behavior and while it is not really documented, I am sure this is treated as semi-public API and the Bundler team are very unlikely to change it. Worst case if they do, is that we lock existing Origen to a Bundler version earlier than the change, then release a new version of Origen to support the new Bundler.
Note that `.bundle/config` should not be committed to revision control since it can contain user-specific paths, though this is the same as before.

Finally, I suspect that setting `site_config.gem_manage_bundler: false` is probably broken, or at least probably doesn't allow the user to fully manage it.
I was wondering whether we should maintain this switch or not, if anyone complains it can be fixed as an incremental update.


lib/origen/site_config.rb
```diff
@@ -1,15 +1,14 @@
 module Origen
   class SiteConfig
+    # NOTE: Gems are not allowed to be required here, only Ruby stlibs
     require 'pathname'
     require 'yaml'
     require 'etc'
     require 'erb'
```
ginty: @coreyeng, note that there were a couple of remaining gems in site_config.rb that I had to remove. I've replicated the red functionality locally, but I couldn't see httparty actually being used anywhere.
This does mean that if you want to add support for a network config file then you will need to use `net/http` rather than a gem to fetch it.

ginty:
@coreyeng, note that there were a couple of remaining gems in `site_config.rb` that I had to remove. I've replicated the `.red` functionality locally, but I couldn't see `httparty` actually being used anywhere.
This does mean that if you want to add support for a network config file then you will need to use `net/http` rather than a gem to fetch it.

ginty:
Note that the sandbox mode works fine with Git repository references from your Gemfile too.

So those who are adventurous enough, could actually start using sandbox mode now with a Git reference to this branch, and your users would be oblivious to anything changing between now and when this is officially released.

pderouen:
Thank you for this update! We're planning to use the archive feature. It's quite a large code difference to read through. I might prefer to kick the tires a bit rather than scour the code. But, I won't have time to do that for a few weeks. I won't have any heart burn if you don't wait for my checkout.

ginty:
I think the diff looks a bit bigger than it is because all the lbin files have been checked in for the first time, you can safely ignore them in any review.

ginty:
OK, this actually works now :-)

Came across an issue after opening the PR where is doesn't work if the system Origen (the one you get when you do origen -v outside of an app) is v0.36.x.

It works fine with 0.35.x and earlier and 0.37.0 has been released today to fix the issue with 0.36.x. Unfortunately there is no way to make it work with 0.36 and so you should make sure that your system is not running that version before trying this.

pderouen:
@ginty is there a command to undo "origen archive --local" ?

pderouen:
Nevermind, I see when it completes it tells you how to undo it. It might be nice to have that info as part of the help text. Or, maybe just add "--unlocal".

ginty:
Thanks for reviewing this @pderouen, added `origen archive --no-local` per your suggestion.

ginty:
I also added the following update to the 'origen new' global command (build a new Origen app) to make it more robust:

- The origen_app_generators plugin is now added to the Origen development environment and is copied into the root of the Origen gem package at release time (so it is now included in an Origen installation)
- If fetching the latest version of origen_app_generators from rubygems.org fails, the packaged version will be used instead
- Any failure to fetch company specific generators will now be failed silently, the rationale being that it is better to offer the user something (the basic templates are now guaranteed to be available), rather than crashing. A power user would have to go in now and debug issues in a problematic environment.

See this commit for details - 
https://github.com/Origen-SDK/origen/pull/305/commits/f70162de8eebdecf19e6a8661aaf1cb5b0a0a52a

pderouen:
Thanks. I was hesitant to try out the archive command not knowing how to undo it.

I just tried out this branch in windows. I get a warning about "./lbin" not being part of my path variable. But, the help text is for linux only. I don't know of a simple way to update the path variable in Windows.

For windows 7:
~~~
(right-click) My Computer -> Properties -> Advanced system settings -> Environment Variables
Path and PATHEXT are under System variables
~~~

Should people just be sent to Google to resolve it?

ginty:
@pderouen, yeah the message comes from here: https://github.com/Origen-SDK/origen/blob/archive/lib/origen/boot/app.rb#L88

Happy to add whatever Windows-specific message you like. Up to you, if you think you know what should be done we should add a helpfule message I think, but if you're not confident in what should be done (like me), then a fallback to a Google is fine.

You make the call...

pderouen:
Well, those are the instructions for Windows 7. I'd have to figure it out for Windows 10... which I would do by asking Google.

ginty:
@coreyeng, @chrisnappi, @priyavadan - any concerns about releasing this?

I would like to get it released and start using it soon - personally I plan to convert all my apps over to local/sandboxed gems mode as soon as this is released as I think it will make the user experience much better.

ginty:
@pderouen, I took a stab at a better Windows check and advice - https://github.com/Origen-SDK/origen/commit/f95c926359d2b3fc295313f3c6649f339e9a4b9a

We can easily tweak in a further release if required once you have tried it out.

PR #308, opened 2018-11-26
RFC5 - Add app/ dir and code generators

ginty:
This PR is a biggie and implements this RFC to add a new app/ dir structure - https://github.com/Origen-SDK/rfcs/blob/master/text/0005-add-app-directory.md

A summary of the changes:

- Adds support for application code in the following dirs:

~~~text
.
├── app/    
│   ├── flows/         // New name for the program dir
│   ├── lib/         
│   ├── parts/         // *NEW* will be discussed at length below
│   ├── patterns/
│   └── templates/
│   
~~~

- Adds an 'origen new' command to generate starter application code, the following generators have been added for now:
  - dut
  - sub_block
  - part
  - model
  - class
  - module

- Any classes defined in files within `app/lib` and `app/parts` will now autoload and users don't have to (and in fact shouldn't) require them manually.
- `dut.reload!` can be used in the console to re-load your target and pick up any files edits that have been made in the meantime (requires that the above autoload was used in the first place though).
- The concept of a definition transaction has been added to parameters, allowing `define_params` blocks to be re-opened and modified. This has mainly been added here to support the behind the scenes loading of the new parts system, but it might be generally useful in other applications.
- `origen_core_support` has been changed over to the new app dir structure to test it and in the process I removed all legacy test code from Origen's `lib/` directory.

These updates are fully backwards compatible and there are no plans to ever remove support for a manually-loaded top-level `lib/` dir. Existing apps can be oblivious to this change, however `origen_app_generators` will be updated to build all new apps with the `app/` dir structure and it will be the recommended way to build apps in future.
The auto-loading feature has not been added to the top-level `lib/` dir to avoid any unintended side-effects on existing applications.

## Changes vs. the RFC

This update definitely implements the spirit of the RFC, though the final `app/` dir structure turned out to be different than originally envisaged.
When writing the RFC, I really wanted to make it much clearer to engineers where to add and find things like pins, parameters, registers, timing, etc., inspired by what I had seen in other tools which made top-level directories for each of these things - like a test program does.

However, once I got into the implementation and starting trying it out, it became clear that this didn't actually work so well in practice for Origen. The reason is that these other tools are built around one application per DUT, whereas Origen is built around the concept of one application supporting many DUTs.
Trying to work out the internal structure of such dirs to make them support multiple-targets, and how to even load and reference them became increasingly convoluted and it just generally began to feel pretty un-Origen-like.
Eventually after some weeks of wrestling with this and getting no where, I decided to change tack and introduced the concept of parts...

### The `app/parts` Directory

A 'part' is a catchall term to cover DUTs and sub-blocks. Honestly, I am not totally sold on the name and if there are other suggestions let's have them. I thought about 'blocks', but thought that that might suggest that sub-blocks were something different, and 'models' didn't really fit because parts are intended to cover both the model and controller.

Here is an example parts dir that contains a couple of DUT parts:

~~~text
.
├── app/    
    ├── parts/        
        ├── dut/          
            ├── derivatives/        
            │   ├── falcon/  
            │   │   ├── derivatives/       
            │   │   ├── controller.rb
            │   │   ├── model.rb
            │   │   ├── parameters.rb
            │   │   ├── pins.rb
            │   │   ├── registers.rb    
            │   │   ├── sub_blocks.rb    
            │   │   └── timesets.rb
            │   └── eagle/
            │       ├── derivatives/              
            │       ├── controller.rb
            │       ├── model.rb
            │       ├── parameters.rb
            │       ├── pins.rb
            │       ├── registers.rb    
            │       ├── sub_blocks.rb    
            │       └── timesets.rb
            ├── controller.rb
            ├── model.rb
            ├── parameters.rb
            ├── pins.rb
            ├── registers.rb    
            ├── sub_blocks.rb    
            └── timesets.rb
~~~

By convention, all parts that are DUTs will live under `app/parts/dut`.

You can generate one of these by running `origen new dut my_dut_name`.

This organizes the code in what should be hopefully a fairly intuitive structure - if you want to add a pin to all of your DUTs, then add it to `parts/dut/pins.rb`, but if you want to add it to only the Falcon for example, then add it to `parts/dut/falcon/pins.rb`.

Additional levels of nesting can be added as deeply as you like, just repeat the same structure, for example within the Falcon's `derivatives/` folder.

The recommended way to do that is to add a namespace or family prefix to the new DUT generator, for example: `origen new dut my_family/my_dut_name`.

That would then give you three layers of hierarchy, dut -> family -> product, in which to work.

It is expected that the majority of a typical application's code will now live in the parts folder. The `app/lib/` folder is for any supporting code which is not really a part. For example, a test program interface would (for now at least) still live in `app/lib/`.

If the model and controller files get too large you can split them up into modules and there is a generator for that and inside the `model.rb` and `controller.rb` files is a comment telling you what command to run to create a module for that particular class.
Generally, you will find the generated files for the pins, etc. have good example comments on how to declare the associated resource (all using existing APIs).

The resource files (`pins.rb`, `registers.rb`, etc.) are loaded automatically whenever an instance of the class defined in `model.rb` is instantiated. In the above example `parts/dut/model.rb` defines class `MyApp::DUT` and `parts/dut/derivatives/falcon/model.rb` defines `MyApp::DUT::Falcon`, which is a child of `MyApp::DUT`. Therefore the same inheritance tree applies for the model and controller classes too - any methods or attributes defined in the parent classes will also be available to their derivative children.

When loading resources, the `parameters.rb` files are all loaded first to allow them to be referenced in any of the other definitions, but after that the load order is undefined, though in all cases the top-level parent files will be loaded first, then each child level in turn.
All of these resource files are evaluated within the context of the model instance, in other words, anything that you write in these files is equivalent to writing it within a method like `def instantiate_regs` within the model.

You can even add your own application-specific files and they will be auto-loaded in the same way, so you could add `fuse.rb` files for example.

Primary sub-blocks, think RAM, ATD, NVM, PLL, etc., should all have their own top-level part and the new sub-block generator will enforce this. i.e. it won't let you run `origen new sub_block atd`, it will make you choose a top-level category and instead do something like `origen new sub_block atd/atd8bit`.

Sub-blocks are not generated with `pins.rb` files since that is a DUT concept, but otherwise they work the same as above.

Finally, there is a generator for a model and controller-less part. Potentially this could be used to define parts that are features, and then these could be composed into models.

All models now have a `load_part` method where you can manually load a part definition to a model. Potentially then in your model's initialize method you could do things like:

~~~ruby
load_part `features/jtag_tap`
load_part `features/ram`
~~~

### The `origen new` Command

Apps now have access to the Origen new command as discussed already, run `origen new -h` for details.

The best way to find out about this is just to try it.

All the generators live in `lib/origen/code_generators/` and their templates are in `templates/code_generators/`. Spec tests have also been added for all of the generators.

This is expected to be a starting point and the community are more than welcome to add more in future.

These should also work well in a future GUI plugin, where they could map directly to menu items to generate new Origen resources within the application.

Note that there is no automated rollback feature, so check everything in before running it and use your revision control system to undo as required.

### The `app/` Directory

The most significant part of the `app/` directory is the parts system that has been discussed at length above.

`app/lib/`, `app/patterns/`, `app/flows/` and `app/templates/` are essentially just re-naming of the original directories from the application top-level and the origen commands like `origen g` and `origen c` have been updated to consider source files from both locations.

### Autoloading

`app/parts/` will autoload files as already discussed, and so to will `app/lib/`.

User's should generally not have to worry about requiring files that they add and the existing Origen new application template used a poor man's auto loading system which simply queried the `lib/` directory for all files and then required them. This was done within the app's top-level `lib/my_app.rb` file.

While that worked OK for smaller apps, I have seen recently that this could be a cause of slow boot up times for large apps since it required the world at boot time.

The new system is probably slower in a race to load the same amount of files, but in practice it should be much quicker since it means that the application will lazy-load.

The auto-load feature is implemented by `lib/origen/loader.rb` and it borrows some of it from Rails, though it doesn't share a whole lot in the end because it is very specific to Origen conventions, i.e. the parts directory and so on.
Basically though, it steps in when Ruby is about to declare a missing constant error, then decomposes the constant name to a file and then loads it if it exists.

To use the auto-loading you simply have to follow the expected naming convention and using the new code generators will ensure that you do that.

### `dut.reload!`

The ability to reload code files in the console fell out naturally from adding the auto load system - basically all classes that are auto-loaded are remembered and then they can be un-defined when `dut.reload!` is called.

This won't work for any classes which have bypassed this system via a manual require, so your mileage may vary with this one.

I suspect that Rails may override Ruby's `require` so that any manually loaded files get refreshed too, but ultimately I wasn't brave enough to try that yet.

### Parameter Transactions

Within the parameters definitions in the parts directory it seems natural to want to modify parameter values per derivative, as well as per parameter set.

So within the parts system, parameters definitions cascade in two directions - what I would call vertical cascading is the existing ability to override with a different parameter set:

~~~ruby
# app/parts/dut/parameters.rb
define_params :default do |params|
  params.my_attribute = 10
end

define_params :probe, inherit: :default do |params, parent|
  params.my_attribute = parent.my_attribute + 5
end
~~~

Additionally, you can now re-open these parameter sets to make horizontal changes between parts, for example if the above was the top-level DUT definition, it is now possible to do this in a child part:

~~~ruby
# app/parts/dut/derivatives/falcon/parameters.rb
define_params :default do |params|
  params.my_attribute = 12
end
~~~

This was previously not allowed since it could have references to other parameter sets, like above where the value for `:probe` is a function of the `:default` value.

Since the part is now loaded within a transaction, all such references are recalculated. The bottom line then is that with the above definitions, the models will behave like this:

~~~ruby
dut = MyApp::DUT.new
dut.params.my_attribute      # => 10
dut.params = :probe
dut.params.my_attribute      # => 15

dut = MyApp::DUT::Falcon.new
dut.params.my_attribute      # => 12         Overridden as you would expect
dut.params = :probe
dut.params.my_attribute      # => 17         The new baseline has been auto-magically applied
~~~

If you find a use for this elsewhere within your app, then wrapping all `define_params` blocks in a transaction as shown below will allow them to be redefined like this:

~~~ruby
Origen::Parameters.transaction do

  # Any define_params in here can be re-opened, though they are frozen
  # once the transaction is complete

end
~~~

## How to Upgrade Existing Apps

Firstly, there is no need to if you don't want to. Whether you wish to do this or not really depends on how much future development work is going to be done on the app and whether the benefits from the new system are worth the churn.

I haven't updated any of my own apps yet, but I would recommend to start with that you play around with the `origen new` command.

This will generate new application code within an `app/` dir and you should see all the auto-loading working.
That will give you a feel for it and then you can decide if it is worth copying all your existing app code across to the new structure or not.


ginty:
I added an additional file to the parts system, `attributes.rb`. This allows DUT and sub-block attributes to be easily defined and managed like this:

~~~ruby
# parts/dut/attributes.rb
@address_width = 12
@data_width = 32
@has_some_bug = false

# parts/dut/derivatives/falcon/attributes.rb
@address_width = 10
@has_some_bug = true

dut = MyApp::DUT.new
dut.address_width      # => 12
dut.data_width         # => 32
dut.has_some_bug?      # => false

dut = MyApp::DUT::Falcon.new
dut.address_width      # => 10
dut.data_width         # => 32
dut.has_some_bug?      # => true
~~~

Regular accessors as well as the `?` style accessors are created whenever the value is a boolean.

All of the `attributes.rb` files are loaded first during a part load, therefore allowing attribute definitions to be referred to in all other resource definition files.

General guidance will be to use attributes whenever you need a different value per product, and parameters if you need a different value per context (e.g. probe, ft, etc.)

pderouen:
@ginty I finally have time to play around a bit with this branch. I don't see any compatibility issues in the apps/plug-ins I've tried.

I'm trying to use my local copy of this branch to create a new app. I don't try to run origen this way very often. Can you (or anybody else) remind me how to get my local origen gem to run at process the new command? Just giving the path to the bin stub in lbin/ doesn't seem to be doing the trick.

ginty:
Hi @pderouen, to answer your specific question, this is the process to setup your global environment to use a specific Origen (you can work this out from the documentation, but it is too verbose I think and needs to be condensed).

Add file `$HOME/.origen/origen_site_config.yml` with the following content:

~~~yaml
user_install_enable: true
~~~

That is your primary switch to turn your custom global environment on and off.

Then run `origen -v` or something in the global scope and you will see it pulling in some gems and stuff.

You will now have a global Gemfile at `$HOME/.origen/Gemfile`, which you can edit as per a normal application gemfile to pull specific versions of Origen or other gems into your global scope.

However, this PR does not make any changes to the global 'origen new' command, since that is implemented mainly by `origen_app_generators`.

Note that references to `origen new` in the above PR info is talking about within an app workspace - so this command behaves differently depending on whether it is invoked inside or outside of an app. Previously it was only available outside an app.

A corresponding PR will be coming to `origen_app_generators` soon, that will add an `app/` dir to the new app builds. It will also simplify it significantly - I think the path, at least for the open source piece of Origen, will be for the new app builder to offer only a pretty simple bare bones app, and now we will have the in-app `origen new` command available for users to quickly add the additional features that they want.

chrishume:
So what's the preferred sequence for converting an old app to the new structure?  I've stumbled around a lot today and still can't get to where I can load the console properly for my new target.  Either I'm missing something required for the auto-load or I'm not understanding whether to call my family of devices duts, parts, or modules.  I always end up with this error:

     uninitialized constant <name of my gem>

I envision this hierarchy for my family of devices:
   <family name>
       --> <first part>
       --> <follow-on parts>

Is the family a dut or a part?  The line between the two seems blurred.  When I do "origen new dut <family>/<first part>", I get an app structure like this:

   app/parts/dut/derivatives/<family>/derivatives/<first part>/model.rb

This doesn't seem as intuitive as the documentation above suggests.  My <family> is a derivative?  Shouldn't <family> be in place of the generic 'dut'?  /app/part and /app/part/dut seem too generic to accurately describe the contents of my gem and its targets.  

I've tried this sequence:
   origen new module <my gem name>
   origen new dut <family>/<first part>
But I still get an uninitialized constant error related to my target/<first part>.rb.

Any idea where I'm going wrong?

chrishume:
Something went wrong with the formatting above.  It was supposed to have a hierarchy of <family>/<first part> and <family>/<follow-on part>.

chrishume:
So what's the preferred sequence for converting an old app to the new structure? I've stumbled around a lot today and still can't get to where I can load the console properly for my new target. Either I'm missing something required for the auto-load or I'm not understanding whether to call my family of devices duts, parts, or modules. I always end up with this error:
 uninitialized constant <name of my gem>

I envision this hierarchy for my family of devices:
  _family_
     --> _first part_
     --> _follow-on part_

Is the family a dut or a part? The line between the two seems blurred. When I do "origen new dut _family_/_first part_", I get an app structure like this:
   app/parts/dut/derivatives/_family_/derivatives/_first part_/model.rb

This doesn't seem as intuitive as the documentation above suggests. My is a derivative? Shouldn't be in place of the generic 'dut'? /app/part and /app/part/dut seem too generic to accurately describe the contents of my gem and its targets.

I've tried this sequence:
origen new module 
origen new dut /

But I still get an uninitialized constant error related to my target/_first part_.rb.

Any idea where I'm going wrong?

chrishume:
Apologies for poor formatting.  Should have said "My family is a derivative? Shouldn't family be in place of the generic 'dut'?"  Then later, "origen new dut family/first part".

ginty:
Hi @chrishume, thanks for trying this out, sorry to hear of the troubles.

A 'part' is a catch-all term to mean either a 'dut', which means a top-level silicon (i.e. a SoC) or a sub-block which is typically an IP within a SoC.
I am definitely open to suggestion about a better name for that, but as discussed already in this thread I can't think of any.

Anyway, I did try generating your scenario and I did find a bug with it - specifically with generating a DUT class within a family like you were doing. I'm not sure if that explains your issue or not but its worth trying again with latest to see if it looks better now.

I've pasted the output from my console session below and it seems to work fine now.

`origen new module` and `origen new class` mean generate a generic Ruby Module/Class (which will end up in app/lib) and so you can probably ignore those for now.

By using `origen new dut` you are saying that you are generating a new top-level model which will be automatically assigned to `dut` whenever you instantiate it within a target and it will be considered the absolute top-level by Origen at runtime.
As such, it has a special importance over a regular sub-block.

Now, part of what this app_dir update is trying to do is to encourage users down a good architecture path.
One of the things that we have learned from experience, is that with DUTs having such a special place, it is often the case in an application that you want to add something to apply to all DUTs.
So the code generator enforces that all DUTs are children of a master DUT class and which is why you don't see your Family at the very top like you might expect.

This is how it should work, it should be very simple:

~~~text
stephen@NXL78024:~/Code/github/origen_core_support$ origen new dut family/first
      create  app/parts/dut/model.rb
      create  app/parts/dut/controller.rb
      create  app/parts/dut/pins.rb
      create  app/parts/dut/timesets.rb
      create  app/parts/dut/parameters.rb
      create  app/parts/dut/registers.rb
      create  app/parts/dut/sub_blocks.rb
      create  app/parts/dut/attributes.rb
      create  app/parts/dut/derivatives/family/model.rb
      create  app/parts/dut/derivatives/family/controller.rb
      create  app/parts/dut/derivatives/family/pins.rb
      create  app/parts/dut/derivatives/family/timesets.rb
      create  app/parts/dut/derivatives/family/parameters.rb
      create  app/parts/dut/derivatives/family/registers.rb
      create  app/parts/dut/derivatives/family/sub_blocks.rb
      create  app/parts/dut/derivatives/family/attributes.rb
      create  app/parts/dut/derivatives/family/derivatives/first/model.rb
      create  app/parts/dut/derivatives/family/derivatives/first/controller.rb
      create  app/parts/dut/derivatives/family/derivatives/first/pins.rb
      create  app/parts/dut/derivatives/family/derivatives/first/timesets.rb
      create  app/parts/dut/derivatives/family/derivatives/first/parameters.rb
      create  app/parts/dut/derivatives/family/derivatives/first/registers.rb
      create  app/parts/dut/derivatives/family/derivatives/first/sub_blocks.rb
      create  app/parts/dut/derivatives/family/derivatives/first/attributes.rb
      create  target/first.rb

New DUT part created, run the following command to select it in your workspace:

  origen t first

stephen@NXL78024:~/Code/github/origen_core_support$ origen i -t target/first.rb
origen(main):001:0> dut
=> <Model/Controller: OrigenCoreSupport::DUT::Family::First:70368519328180/OrigenCoreSupport::DUT::Family::FirstController:70368516049340>
~~~

chrishume:
@ginty Okay, that makes sense.  Perhaps the "dut" in app/parts/dut could be named "master_dut" to make that special place in the architecture more apparent?

Still get the same error, though.  I'll try to catch you on IM.

redxeth:
@ginty 
'parts' alternatives...
- models
- ips
- ipblocks
- ipmodels

I know you said no to 'models' above-- but looking at RoR they have app/models diras well, and i think you can have a models dir that contains both model(s) and their respective controllers.

Just to try how this looks:
~~~
├── app/    
│   ├── models/
│       ├── dut/
│       ├── ipblockA/
│        |   ├── model.rb
│        |   ├── controller.rb
│       └── ipblockB/
│   
~~~

or possibly better:

~~~
├── app/    
│   ├── ipBlocks/
│       ├── dut/
│       ├── ipBlockA/
│        |   ├── model.rb
│        |   ├── controller.rb
│       └── ipBlockB/
│   
~~~

ginty:
but also `app/controllers`.

That's the direction I started off with here, but you ended up with deep nesting replicated in each dir type and I thought for Origen code is was better just to have one place where the hierachy was laid out and then from in there easy access to the model/controller/regs etc. that exist in the same level.

ginty:
but anyway, yeah I think models is probably the more natural Origen name for this, then we just kinda overlook that it also contains the model's controllers too.

redxeth:
@ginty well they are 'model controllers' technically-- the serve the model, like a helper.

ginty:
@redxeth, @chrishume, I updated this to rename `app/parts` to `app/models` per earlier discussion in this thread.
There is no backwards compatibility added to enable the use of either name, so when you update your existing apps to the new version of this branch you will need to rename your directory accordingly.

I'm also preparing to release this soon, so you or anyone else monitoring this thread please shout soon if you see any things that should be changed.

Thanks!

redxeth:
@ginty still need me to help do some documentation for this branch?

redxeth:
Hmm, updating my project app to latest app_dir commit breaks something.  Works ok with commit 'f178713' (my merge from Feb. 1).  Looking into it.

redxeth:
Nevermind.  Assumed something incorrectly.  All is good after dirname change.

chrishume:
The latest updates work on my testcases as well.

redxeth:
@ginty Looks like the doc changes are still needed, particularly for the guides/starting/architecture page, to include the right directories there.  Assume that we'd go full force recommending the new directory structure and lose any references to the old one (though it's still compatible).  Or would that alienate users of the 'old' structure?

ginty:
@redxeth, any documentation contribution would be very welcome.
Yes, we'd write the docs to only consider/recommend with maybe just a warning note mentioning that original Origen applications used a different structure and that they might see this in the videos.

chrishume:
@ginty FYI when you do 'origen new dut/model/etc.' you get the usual Ruby crash dump instead of a warning that no name was supplied for the dut/model/whatnot that you were trying to create.  We might want to add that warning in case anyone (like me) doesn't put 2 and 2 together very quickly.

Missing dut/model/etc. name:
COMPLETE CALL STACK
-------------------
undefined method `split' for nil:NilClass
...

With dut/model/etc. name:
origen new dut ref_dut

      create  app/models/dut/model.rb
      create  app/models/dut/controller.rb
...

redxeth:
@chrishume BTW, took me a while to remember how gem building worked-- but in addition to renaming the app/parts directory, you need also to modify you gemspec file to include `app/models` instead of `app/parts`, or your released gem won't work.  Ask me how I know...

redxeth:
@ginty this will require an update to code generator app_dir branch-- let me do that.

ginty:
> @ginty FYI when you do 'origen new dut/model/etc.' you get the usual Ruby crash dump instead of a warning that no name was supplied for the dut/model/whatnot that you were trying to create. 

OK, I've fixed this @chrishume, thanks for reporting it.

ginty:
The `app/models` directory has been renamed to `app/blocks`, it became clear when writing documentation that we need to have a unique name for this new concept of an entity that covers the model/controller and all other components of the Origen representation of a silicon component.

PR #309, opened 2018-11-26
msb0 Bit order bug fixes

pderouen:
Some clean up of msb0 functionality. @ginty my updates caused the scan register spec tests to fail when accessing .data of a port object. I left some comments in the code to explain what I think was happening. I will remove the extra comments before merging.

### DIFF
--- a/lib/origen/ports/port.rb
+++ b/lib/origen/ports/port.rb
@@ -68,7 +68,7 @@ module Origen
         end
       end
 
-      # Not sure if this is the right way to fix the infinite loop
+      # Prevent infinite loop if a child bit collection checks bit_order
       def bit_order
         parent.bit_order
       end
--- a/lib/origen/ports/port.rb
+++ b/lib/origen/ports/port.rb
@@ -68,6 +68,11 @@ module Origen
         end
       end
 
+      # Not sure if this is the right way to fix the infinite loop
+      def bit_order
+        parent.bit_order
+      end
+
       def drive(value = nil, options = {})
         value, options = nil, value if value.is_a?(Hash)
         if options[:index]
--- a/lib/origen/registers/bit_collection.rb
+++ b/lib/origen/registers/bit_collection.rb
@@ -284,16 +284,9 @@ module Origen
       #   reg(:control).data         #  => 0x55, assuming the reg has the required bits to store that
       def data
         data = 0
-        if bit_order == :msb0
-          reverse_shift_out_with_index do |bit, i|
-            return undefined if bit.is_a?(Origen::UndefinedClass)
-            data |= bit.data << i
-          end
-        else
-          shift_out_with_index do |bit, i|
-            return undefined if bit.is_a?(Origen::UndefinedClass)
-            data |= bit.data << i
-          end
+        shift_out_with_index do |bit, i|
+          return undefined if bit.is_a?(Origen::UndefinedClass)
+          data |= bit.data << i
         end
         data
       end
@@ -309,16 +302,9 @@ module Origen
       # Returns the reverse of the data value held by the collection
       def data_reverse
         data = 0
-        if bit_order == :msb0
-          shift_out_with_index do |bit, i|
-            return undefined if bit.is_a?(Origen::UndefinedClass)
-            data |= bit.data << i
-          end
-        else
-          reverse_shift_out_with_index do |bit, i|
-            return undefined if bit.is_a?(Origen::UndefinedClass)
-            data |= bit.data << i
-          end
+        reverse_shift_out_with_index do |bit, i|
+          return undefined if bit.is_a?(Origen::UndefinedClass)
+          data |= bit.data << i
         end
         data
       end
@@ -345,14 +331,8 @@ module Origen
         end
         value = value.data if value.respond_to?('data')
 
-        if bit_order == :msb0
-          size.times do |i|
-            self[i].write(value[size - i - 1], options)
-          end
-        else
-          size.times do |i|
-            self[i].write(value[i], options)
-          end
+        size.times do |i|
+          self[i].write(value[i], options)
         end
         self
       end
--- a/lib/origen/registers/bit_collection.rb
+++ b/lib/origen/registers/bit_collection.rb
@@ -284,9 +284,16 @@ module Origen
       #   reg(:control).data         #  => 0x55, assuming the reg has the required bits to store that
       def data
         data = 0
-        shift_out_with_index do |bit, i|
-          return undefined if bit.is_a?(Origen::UndefinedClass)
-          data |= bit.data << i
+        if bit_order == :msb0
+          reverse_shift_out_with_index do |bit, i|
+            return undefined if bit.is_a?(Origen::UndefinedClass)
+            data |= bit.data << i
+          end
+        else
+          shift_out_with_index do |bit, i|
+            return undefined if bit.is_a?(Origen::UndefinedClass)
+            data |= bit.data << i
+          end
         end
         data
       end
@@ -302,9 +309,16 @@ module Origen
       # Returns the reverse of the data value held by the collection
       def data_reverse
         data = 0
-        reverse_shift_out_with_index do |bit, i|
-          return undefined if bit.is_a?(Origen::UndefinedClass)
-          data |= bit.data << i
+        if bit_order == :msb0
+          shift_out_with_index do |bit, i|
+            return undefined if bit.is_a?(Origen::UndefinedClass)
+            data |= bit.data << i
+          end
+        else
+          reverse_shift_out_with_index do |bit, i|
+            return undefined if bit.is_a?(Origen::UndefinedClass)
+            data |= bit.data << i
+          end
         end
         data
       end
@@ -331,8 +345,14 @@ module Origen
         end
         value = value.data if value.respond_to?('data')
 
-        size.times do |i|
-          self[i].write(value[i], options)
+        if bit_order == :msb0
+          size.times do |i|
+            self[i].write(value[size - i - 1], options)
+          end
+        else
+          size.times do |i|
+            self[i].write(value[i], options)
+          end
         end
         self
       end
--- a/lib/origen/registers/msb0_delegator.rb
+++ b/lib/origen/registers/msb0_delegator.rb
@@ -1,0 +1,28 @@
+module Origen
+  module Registers
+    require 'delegate'
+
+    # Thin wrapper around pin objects to modify bit number interpretation
+    #
+    # This is provided as a convenience to make user code more readable
+    class Msb0Delegator < ::Delegator
+      def initialize(reg_object)
+        @reg_object = reg_object
+      end
+
+      def __getobj__
+        @reg_object
+      end
+
+      def __object__
+        @reg_object
+      end
+
+      def inspect
+        @reg_object.inspect with_bit_order: true
+      end
+    end
+    
+  end
+end
+  
--- a/lib/origen/registers/msb0_delegator.rb
+++ b/lib/origen/registers/msb0_delegator.rb
@@ -19,7 +19,7 @@ module Origen
       end
 
       def inspect
-        @reg_object.inspect with_bit_order: true
+        @reg_object.inspect with_bit_order: :msb0
       end
     end
     
--- a/lib/origen/registers/reg.rb
+++ b/lib/origen/registers/reg.rb
@@ -137,7 +137,7 @@ module Origen
       end
 
       def inspect(options = {})
-        # options[:with_bit_order] = true
+        options[:with_bit_order] = true
         wbo = options[:with_bit_order] || false
         domsb0 = wbo && bit_order == :msb0
         dolsb0 = !domsb0
@@ -205,13 +205,13 @@ module Origen
         num_bytes.times do |byte_index|
           # Need to add support for little endian regs here?
           byte_number = num_bytes - byte_index
-          if true # dolsb0
+            # if true # dolsb0
             max_bit = (byte_number * 8) - 1
             min_bit = max_bit - 8 + 1
-          else
-            min_bit = (byte_index * 8)
-            max_bit = min_bit + 7
-          end
+          # else
+          #   min_bit = (byte_index * 8)
+          #   max_bit = min_bit + 7
+          # end
 
           # BIT INDEX ROW
           line = '  '
@@ -601,24 +601,24 @@ module Origen
         @lookup.each { |_k, v| split_bits = true if v.is_a? Array }
 
         if split_bits == false
-          if true # bit_order == :lsb0
+            # if true # bit_order == :lsb0
             current_pos = size
-          else
-            current_pos = 0
-          end
+          # else
+          #   current_pos = 0
+          # end
           # Sort by position
           @lookup.sort_by { |_name, details| -details[:pos] }.each do |name, details|
-            if true # bit_order == :lsb0
+              # if true # bit_order == :lsb0
               pos = details[:bits] + details[:pos]
-            else
-              pos = details[:pos]
-            end
+            # else
+            #   pos = details[:pos]
+            # end
             if options[:include_spacers] && (pos != current_pos)
-              if true # bit_order == :lsb0
+                # if true # bit_order == :lsb0
                 collection = BitCollection.dummy(self, nil, size: current_pos - pos, pos: pos)
-              else
-                collection = BitCollection.dummy(self, nil, size: pos - current_pos, pos: current_pos)
-              end
+              # else
+              #   collection = BitCollection.dummy(self, nil, size: pos - current_pos, pos: current_pos)
+              # end
               unless collection.size == 0
                 if block_given?
                   yield nil, collection
@@ -638,18 +638,18 @@ module Origen
                 result << [name, collection]
               end
             end
-            if true # bit_order == :lsb0
+              # if true # bit_order == :lsb0
               current_pos = details[:pos]
-            else
-              current_pos = details[:bits] + details[:pos]
-            end
+            # else
+            #   current_pos = details[:bits] + details[:pos]
+            # end
           end
           if options[:include_spacers] && (current_pos != 0)
-            if true # bit_order == :lsb0
+              # if true # bit_order == :lsb0
               collection = BitCollection.dummy(self, nil, size: current_pos, pos: 0)
-            else
-              collection = BitCollection.dummy(self, nil, size: size - current_pos, pos: current_pos)
-            end
+            # else
+            #   collection = BitCollection.dummy(self, nil, size: size - current_pos, pos: current_pos)
+            # end
             unless collection.size == 0
               if block_given?
                 yield nil, collection
--- a/lib/origen/registers/reg.rb
+++ b/lib/origen/registers/reg.rb
@@ -1,4 +1,5 @@
 require 'json'
+require 'origen/registers/msb0_delegator'
 module Origen
   module Registers
     # The register class can be used to represent not only h/ware resgisters,
@@ -95,6 +96,12 @@ module Origen
         options.each_value { |bit_desc| bit_desc[:pos] = @size - bit_desc[:pos] - bit_desc[:bits] } if bit_order == :msb0
 
         add_bits_from_options(options)
+
+        @msb0_delegator = Msb0Delegator.new(self)
+      end
+
+      def msb0
+        @msb0_delegator
       end
 
       # Returns the bit order attribute of the register (either :msb0 or :lsb0). If
@@ -137,7 +144,6 @@ module Origen
       end
 
       def inspect(options = {})
-        options[:with_bit_order] = true
         wbo = options[:with_bit_order] || false
         domsb0 = wbo && bit_order == :msb0
         dolsb0 = !domsb0
--- a/lib/origen/registers/reg.rb
+++ b/lib/origen/registers/reg.rb
@@ -90,6 +90,14 @@ module Origen
         @size.times do |n|
           @bits << Bit.new(self, n, writable: @init_as_writable, undefined: true)
         end
+
+        # Internally re-map msb0 register descriptions as lsb0
+        if @bit_order == :msb0
+          options.each_key do |bit_id|
+            options[bit_id][:pos] = @size - options[bit_id][:pos] - options[bit_id][:bits]
+          end
+        end
+
         add_bits_from_options(options)
       end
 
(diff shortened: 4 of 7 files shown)


ginty:
It seems like this is modifying `BitCollection#shift_out` to make it MSB/LSB aware, when that is intentionally meant to be a data shift rather than a physical shift.

See - https://origen-sdk.org/origen/guides/models/registers/#Bit_Order_and_Serial_Processing

Could the downstream problem not be solved here by using `shift_out_left` instead, which is already MSB/LSB aware?

ginty:
Or maybe it's modifying the right/left versions to make them un-aware.

In any case, we seem to already have APIs to select whether you want an MSB/LSB dependent shift or not, so I'm confused as to why a change is need here?

ginty:
Also surely bit[31] should always be bit[31] of the data value regardless of the physical placement? This is intended to be a logical rather than a physical API. 😕

pderouen:
Apparently, I misunderstood the intention of some of the API's. If you run my new rspec test in an unmodified workspace, you'll see our issue.

pderouen:
In our msb0 projects if you want to set bit 0 of a 32-bit register, you'd write 0x80000000 to it. So, I'd expect reg.bit[0].write 1 to result in a data value of 0x80000000.

ginty:
Well, the bottom line is that I don't really know how an msb0 register should behave.
However, I do know that the current API and behavior was fleshed out and reviewed by stakeholders in an RFC - https://github.com/Origen-SDK/origen/pull/57

So, we should not be fundamentally changing the behavior here in a quick PR without fully thinking it through.

ginty:
I would also question whether you have to write that 0x8000_0000 because that is the true behavior of the register, or is it because you are writing it through several layers of conversion and it gets switched around somewhere along the way?

That is the exact opposite of the discussed and signed off behavior when this was implemented, will another project want to switch back again?

To me, changing data like that is conventionally the job of the endianness system, not the bit order.
And btw, I would be the first to admit that Origen does not have a good story on the endianness part today, but I had thought until now that at least we had the bit order side of things resolved.

pderouen:
It is the true behavior of the register.

pderouen:
As far as I know, we are currently on our first attempt at actually using the :msb0 feature. In the past we've simply not taken electronic delivery of registers. When manually defining registers, we'd define them as :lsb0 and interact only with the named fields. 

When I take the same register definition and in :msb0 format (which I hate with a passion) several methods are not behaving correctly in my opinion. This isn't intended to be a rushed PR, perhaps not 100% correct, but not rushed.

chrisnappi:
Just to comment that Paul is not crazy :) There are issues here that Ronnie has worked to patch up - but it is not surprising there are still outstanding issues.

pderouen:
msb0 will drive me crazy. After >10 years of exposure to it, I still can't believe anybody does it that way.

ginty:
OK, I have no problem with you all sorting this out the way it should be.

I guess I was just making the point that this is a breaking API change and it needs reviewed and signed off much wider than just by me, and the docs need updated, etc.

Does anyone have a feeling for whether this is an industry standard convention, or just NXP?

pderouen:
The story that I'm told is that this actually came from IBM during consortium days of PPC

pderouen:
If I've misunderstood the expected behavior of shift out right and left, I will un-break them. It was only .data and .copy methods causing issues for us.

pderouen:
@ra2727dude, @chrishume any comments on the expected behavior of .data and .copy functions when :msb0 is the bit order?

pderouen:
@chrishume can we start a petition to end use of msb0? I'll sign.

ginty:
Just throwing it out there for consideration - should Origen abstract the fact that some registers are MSB0 from a user perspective?

chrishume:
@ginty I would likely be in favor of that.  You're effectively suggesting Origen be a (default) LSB0-only interface, correct?

pderouen:
Just take in the msb0 definition as an lsb0 register? Fine by me as long as we can make the docs that get build clarify how bits are accessed. We actually do this is some of our reference manuals.

ginty:
Something like that, can't say I've really thought through how, but conceptually if this is something that you all hate, why don't we make Origen make the problem go away and add a nice marketing feature in the process.
So something like all regs behave like LSB0 from a user perspective, probably at least set a small flag or something to indicate when regs are MSB0 in the console and web views, and then just deal with all the transformation weirdness under the hood.

pderouen:
If other msb0 folks agree, I'd be happy to change this PR accordingly.

pderouen:
maybe would need to add a "doc_label" method to access the bit based on the name that would be in an RM and add a row to the console and web docs display to clarify?

chrisnappi:
Using bit field value ids would be great.  Assuming correct XML.

pderouen:
So, .bit would access by bit field value. I'm trying to think of a clean concise name for accessing based on the bit field label.

~~~ruby
  reg.bit(by_label: true)[0].write 1   # something like that?
~~~

Then the docs and debug display would have a row for bit value and a second for label (if it's :msb0)?

When an :msb0 reg is defined, Origen would internally remap it to an :lsb0 register and jumble/unjumble based on how you choose to access it. Something like that?

ginty:
~~~ruby
reg.bit[0]
~~~

is a method call to a BitCollection object - https://github.com/Origen-SDK/origen/blob/master/lib/origen/registers/bit_collection.rb#L44

So, weirdly, I think Ruby would allow this:

~~~ruby
reg.bit[0, by_label: true]
~~~

chrishume:
Sounds like we might need to expand :msb0 coverage for CrossOrigen. For reference, here's the IEEE IP-XACT handling of big
ipxact_big_endian
 endian:

pderouen:
Is that an msb0 specific behavior or just another variant of endianess with some msb0 spice on top?

chrishume:
No objection on my end - the copy and data operations above look correct.  

Separately, we would need to make sure the XML side is handled correctly by CrossOrigen, I would think (to avoid losing the :msb0 nature of specific registers/sections of a memory map).

pderouen:
@rlaj, @chrisnappi, @ra2727dude, any comment?

Do you agree that this (https://github.com/Origen-SDK/origen/pull/309#issuecomment-442099966) is how .data and .copy should work?

Would you be happy with this (https://github.com/Origen-SDK/origen/pull/309#issuecomment-442146378) change?

ra2727dude:
It looks good to me, I only have the example that William has shown us which breaks in the stam_driver reg.copy command, which I believe this solves.

Would the :by_label option may be needed in older PPC apps if they take this update ? Maybe they did not use MSB0 ?

-joel

pderouen:
Yes, this would fix the issue William is seeing (data winding up in the wrong place when doing a copy_all from an msb0 register).

Our older apps just defined registers as lsb0. So, wouldn't have any issue taking this update. In any case, presently bit[0] does not reference the msb of an msb0 register. So, I think the proposed change has better backwards compatibility than what is in the PR at the moment.

pderouen:
@ginty after reading the guide: 

https://origen-sdk.org/origen/guides/models/registers/#Bit_Order_and_Serial_Processing

I agree that shift out left and right should be deprecated. The msb0 example should look like this (if it's supposed to be the equivalent of the lsb0 example):

~~~
0x24 - :ctrl
  ================================================================================================================
  |     0       |     1       |     2       |     3       |     4       |     5       |      6      |      7      |
  |             |             |             |             |             |             |             |             |
  |             |             |             |             |             |             |             |             |
  ----------------------------------------------------------------------------------------------------------------
  |      8      |      9      |      10     |      11     |      12     |      13     |      14     |      15     |
  |    :coco    |    :aien    |    :diff    |                             :adch[4:0]                              |
  |    0(RO)    |      0      |      0      |                                0x1F                                 |
  ----------------------------------------------------------------------------------------------------------------

O~~~

ginty:
@pderouen, actually I think it is the `shift_out` and `reverse_shift_out` methods that should either be deprecated or just made aliases of `shift_out_left/right`. The former methods were added specifically to support the LSB0/MSB0 differences and the latter are original methods which are used in all serial port drivers today.

ginty:
For API inspiration, also consider how pin functions work.
Depending on which function reference you use to get handle on the pin determines how it will behave:

~~~ruby
add_pin :pin1

pin(:pin1).add_function :nvm_fail, direction: :output
pin(:pin1).add_function :tdi, direction: :input

# Returns the same underlying pin object
pin(:nvm_fail).id        # => :pin1
pin(:tdi).id             # => :pin1

# But it automagically behaves differently
pin(:nvm_fail).direction        # => :output
pin(:tdi).direction             # => :input
~~~

I struggle a little bit with how to name this bit order stuff, but perhaps something like this would work well:

~~~ruby
# We have a simple register, defined with MSB0 bit order
dut.add_reg :my_reg, 0, size: 16, bit_order: :msb0

# By default then, Origen will make this look as if it is a regular MSB0 register
dut.my_reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │                                                    d[15:8]                                                    │
  │                                                      0x0                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │                                                    d[7:0]                                                     │
  │                                                      0x0                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# And it will behave like one
dut.my_reg.write(0x5500)
dut.my_reg[0].write(1)

dut.my_reg.data   # => 0x5501
dut.my_reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │                                                    d[15:8]                                                    │
  │                                                     0x55                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │                                                    d[7:0]                                                     │
  │                                                     0x01                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# If we call .with_bit_order on the register, it will return the register with an invisible wrapper around it which makes it bit-order aware
dut.my_reg.with_bit_order.data   # => 0x80AA
dut.my_reg.with_bit_order   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │      0      │      1      │      2      │      3      │      4      │      5      │      6      │      7      │
  │                                                    d[0:7]                                                     │
  │                                                     0x80                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      8      │      9      │     10      │     11      │     12      │     13      │     14      │     15      │
  │                                                    d[8:15]                                                    │
  │                                                     0xAA                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# This .with_bit_order method would enable bit order aware behavior for everything, and obviously
# for LSB0 registers it would have no effect
dut.my_reg.reset
reg = my_reg.with_bit_order
reg.write(0x5500)
reg[0].write(1)

reg.data   # => 0x5501
reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │      0      │      1      │      2      │      3      │      4      │      5      │      6      │      7      │
  │                                                    d[0:7]                                                     │
  │                                                     0x55                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      8      │      9      │     10      │     11      │     12      │     13      │     14      │     15      │
  │                                                    d[8:15]                                                    │
  │                                                     0x01                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘
~~~

I think the intention though, would be that users would never use this `.with_bit_order` method and would just interact with the register per its MSB0 Origen representation.

Register access protocol drivers would be where this would be used to align things up with the physical h/ware.

The key question though is does that make things easier for users, or does trying to abstract the h/ware like that actually make it more difficult?

pderouen:
The example doesn't make sense to me. The data should not change with bit order. When a register is defined as msb0, it just means that bit 0 references the msb.

pderouen:
So, reg with order should only have the bit labels in the display changing

pderouen:
https://en.wikipedia.org/wiki/Bit_numbering

pderouen:
Maybe bit_order wasn't the best name for the attribute. bit_label_order would have been less confusing. :-|

ginty:
OK, got it @pderouen, that makes more sense.

So, the `with_bit_order` API (if we were to go that way), would work like this:

~~~ruby
# We have a simple register, defined with MSB0 bit order
dut.add_reg :my_reg, 0, size: 16, bit_order: :msb0

# By default then, Origen will make this look as if it is a regular MSB0 register
dut.my_reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │                                                    d[15:8]                                                    │
  │                                                      0x0                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │                                                    d[7:0]                                                     │
  │                                                      0x0                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# And it will behave like one
dut.my_reg.write(0x5500)
dut.my_reg[0].write(1)

dut.my_reg.data   # => 0x5501
dut.my_reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │                                                    d[15:8]                                                    │
  │                                                     0x55                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │                                                    d[7:0]                                                     │
  │                                                     0x01                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# If we call .with_bit_order on the register, it will return the register with an invisible wrapper
# around it which makes it bit-order aware
dut.my_reg.with_bit_order.data   # => 0x5501
dut.my_reg.with_bit_order   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │      0      │      1      │      2      │      3      │      4      │      5      │      6      │      7      │
  │                                                    d[15:8]                                                    │
  │                                                     0x55                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      8      │      9      │     10      │     11      │     12      │     13      │     14      │     15      │
  │                                                    d[7:0]                                                     │
  │                                                     0x01                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

# This .with_bit_order method would enable bit order aware behavior for everything, which
# basically means that any numeric references to a bit position would return different bits:
dut.my_reg.reset
reg = dut.my_reg
reg_with_bit_order = dut.my_reg.with_bit_order

reg_with_bit_order[0..3].write(0x5)
reg_with_bit_order.data     # => 0x5000
reg_with_bit_order[0..3].data     # => 5
reg_with_bit_order[15..12].data   # => 0   

reg.data           # => 0x5000
reg[0..3].data     # => 0
reg[15..12].data   # => 5   


reg   # =>

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │     15      │     14      │     13      │     12      │     11      │     10      │      9      │      8      │
  │                                                    d[15:8]                                                    │
  │                                                     0x50                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      7      │      6      │      5      │      4      │      3      │      2      │      1      │      0      │
  │                                                    d[7:0]                                                     │
  │                                                      0x0                                                      │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘

reg_with_bit_order  # =>  

  ╒═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╤═════════════╕
  │      0      │      1      │      2      │      3      │      4      │      5      │      6      │      7      │
  │                                                    d[15:8]                                                    │
  │                                                     0x50                                                      │
  ├─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┼─────────────┤
  │      8      │      9      │     10      │     11      │     12      │     13      │     14      │     15      │
  │                                                    d[7:0]                                                     │
  │                                                      0x0                                                       │
  └─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┴─────────────┘
~~~

pderouen:
As they say where I'm from, "Mais, dats mo' betta yeah!"

ginty:
Yep, sounds good @pderouen

pderouen:
I will continue with updates to correct failing tests and add the additional functionality discussed. But, I wanted to show the :msb0 register description re-map to :lsb0 format. Any feedback before I get too far down the road?

ginty:
@pderouen, did you mean to show something?

pderouen:
It's in the diffs. The re-mapping from :msb0 to :lsb0 when a register is created.

chrisnappi:
I have been quiet on this because I am torn, not because I don't care.

I absolutely despise :msb0 - remapping to :lsb0 _seems_ like a nice idea.  My concerns revolve around not catching corner cases in the ecosystem leading to massive confusion and in confusion when comparing to other documentation sources.  For instance we recently had trim documentation stating things like 'program trim[1:3] to 0x1'.  What happens when I say reg.bits[:trim]=0x1?

pderouen:
@chrisnappi in that case the bit field named trim will be set to a value of 1. The difference would be when you access the bit fields by bit number. That's where the potential confusion comes in.

chrisnappi:
:) Confusion already.  What will trim[1] and trim[3] be?

pderouen:
I agree about potential confusion. That's why I'm proposing the updates to the debug display and web docs generation.

But, 1 plus (at least for us) is that re-mapping to :lsb0 would allow better reuse. reg.bit[0] would always refer to the lsb. If you needed to reference a bit by the label you could. Hopefully if the doc updates are done right confusion can be avoided.

Maybe I could also add a warning message when an msb0 register is accessed by bit number to remind folks of the behavior?

pderouen:
It depends how you access the register. In your case you'd want to do this to ensure you get the correct value in the field:

~~~ruby
  dut.reg.with_bit_order.bits(:trim)[1..3].write 1
~~~

I don't think origen is injecting the confusion though because the first question you ALWAYS have to ask with stinkin' msb0 is "what do you mean?" any time you refer to a bit number. Half the time people are actually referring to the bits by lsb0 format rules.

pderouen:
I think the re-map to :lsb0 is the right solution to get registers defined in :msb0 format to behave correctly in origen.

The question about user interface and avoiding additional confusion is valid. Perhaps the question is whether the default way of accessing bits by bit number should be :lsb0 or the native ordering of the register when it was defined.

pderouen:
something like this for the debug console?
image

priyavadan:
@pderouen the CI builds are failing with respect to the bit ordering, do the spec tests need to be updated as well?

In terms of msb0 and lsb0 discussion above, i am still trying to conclude if going this way is going to result in confusion for legacy applications.

pderouen:
Yeah, I'm still making updates. The CI builds will fail until I'm done.

Does anybody have legacy apps that actually use msb0? We do not. Any apps that were using it previously I'd expect to either be kind of hacky kludges or working on the full register (which would still work).

pderouen:
@priyavadan I tried to stop by a minute ago. What specifically were to thinking would be confusing?

priyavadan:
@pderouen i am in a planning summit right now, will stop by after or we can discuss tomorrow as well!

pderouen:
I've left this PR open to continue collecting input. I'll probably close it soon and reopen when the new updates are complete. But, after thinking on this some some, perhaps something like this would be better:

~~~ruby
  # any register can be accessed explicitly with msb0 or lsb0 bit numbering like this:
  
  # explicitly tell Origen that you're referring to the LSB when you specify bit 0
  dut.reg(:some_reg).lsb0.bits(:some_field)[0]

  # explicitly tell Origen that you're referring to the MSB when you specify bit 0
  dut.reg(:some_reg).msb0.bits(:some_field)[0]

  # display the register with either numbering to the display console the same way
  (byebug) dut.reg(:some_reg).lsb0       # display with lsb0 numbering

  (byebug) dut.reg(:some_reg).msb0    # display with msb0 numbering


  # The default interpretation of the bit numbers will be:
  #    whatever Origen does currently (lsb0 - I think, will confirm)
  # Origen will expose to change this default if desired (somebody somewhere prefers msb0)
  Origen.bit_number_interpretation = :reg_def_order
  Origen.bit_number_interpretation = :lsb0
  Origen.bit_number_interpretation = :msb0
~~~

If the register is defined using msb0 numbering, the web docs will show both bit numbering schemes to avoid doc mismatch confusion.

ginty:
Looks nice to me

ginty:
@pderouen, just capturing our off-line discussion for the benefit of others:
probably you should not have a global setting for this, because then you could break register updates in 3rd party plugins, etc.
I think a higher level control is a good idea, but I would limit it to model-level, i.e. setting msb0 on a model would apply to all of its child registers, but wouldn't go sideways.
I guess, its an open question of whether it should cascade or not? I think it should, though that does mean that setting it on a DUT model could break 3rd party sub-blocks that you could instantiate.
I don't think Origen should be safe at the expense of usability for responsible users though, and I can imagine many scenarios where it would be good to be able to set this at the top-level of your particular domain and have it cascade down.

pderouen:
I just confirmed that the currently released Origen version interprets bit numbers as lsb0 even if the register is defined using msb0.

~~~ruby
  reg :some_reg, 1, size: 32, bit_order:  :msb0 do |r|
    r.bits 31..0, :df
  end

  dut.some_reg[0].write 1
  dut.some_reg.data                        # => 1
~~~

ginty:
Yep, I think all it does right now is change the order of the `shift_out` and `reverse_shift_out` methods

pderouen:
I'm going to close this PR now. Contact me directly if you have any more input or concerns. I hope to submit a new PR for this next week.

PR #310, opened 2018-11-30
Added site config broadcast system

ginty:
A small update to add the ability to output broadcast messages (shown to all users every time they invoke Origen) from site configs.

The given text it just output verbatim to the console and can be optionally colored, the example output is shown below.

The intended use case within NXP is to allow us to deprecate Origen/Ruby installations and advise users to transition to a supported installation.

~~~text
stephen@xwing:~/Code/github/origen$ origen -v

These lines will be output to the console in regular text

These lines will be output to the console in amber text

These lines will be output to the console in red text

These lines will be output to the console in green text

Application: 0.38.0
     Origen: 0.38.0
~~~

### DIFF
--- a/bin/origen
+++ b/bin/origen
@@ -182,6 +182,29 @@ else
 end
 
 begin
+  require 'colored'
+  # Emit all broadcast messages before executing all commands
+  if Origen.site_config.broadcast_info
+    puts
+    Array(Origen.site_config.broadcast_info).each { |line| puts line }
+    puts
+  end
+  if Origen.site_config.broadcast_warning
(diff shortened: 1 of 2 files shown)


ginty:
It's OK in this case because it occurs after this line - https://github.com/Origen-SDK/origen/blob/master/bin/origen#L172

After that the application/user/global bundle is loaded, and as colored is a hard dependency of Origen it will definitely be available.
It's only early in the boot process where the site config is first loaded and it is trying to work out where the gems should live etc. that you need to watch out.

PR #318, opened 2018-12-27
Linux subsystem for windows

welguisz:
Instructions for getting Origen to run on the Windows Subsystem for Linux.  Most of the issues is just getting the base installation running, then just using 'Linux instructions'

### DIFF
--- a/templates/web/guides/starting/installing.md.erb
+++ b/templates/web/guides/starting/installing.md.erb
@@ -5,6 +5,9 @@ Origen can be installed on Linux or Windows, this guide refers to the Linux inst
 If running on windows is of interest then please see this guide instead:
 [How to Install Origen on Windows](<%= path "guides/starting/installwin" %>)
 
+If you want to run on Linux Subsystem for Windows then please see this guide instead:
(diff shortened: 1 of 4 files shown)


info-rchitect:
@welguisz Thanks for the PR.  I will get to trying this out by the weekend.

priyavadan:
LGTM!

ginty:
I'm not sure about this one.

To me, how to install Origen in the Windows Linux subsystem, is exactly the same as how to install Origen on Linux.
So this guide is more about how to install the Linux subsystem, but I don't think that is our responsibility and something that we should be maintaining a guide for.

The Linux install guide does mention installing Ruby, but the detail is handled via a link to a 3rd party.

I would far rather we take the same approach here, basically just adding a line to the Windows install guide that says:

> We recommend that when running on Windows 10 the Linux subsystem is used to run Origen, here is a link to a guide on how to install it - <link to a good 3rd party guide>. After that you can install Origen via the instructions for Linux.

ginty:
There is an official guide from Microsoft which looks good to me - https://docs.microsoft.com/en-us/windows/wsl/install-win10

ginty:
Closing this per my earlier comments and because the contributor is no longer involved with Origen.

PR #321, opened 2018-12-31
Register export with flexibel path to subbblocks

williamcheong:
Exported ruby register definition file can point to the vendor directory under Origen.root instead of the hard path to particular directory under specific user which not accessible by other users. Changes on exporter.rb:
1. Add options[:dir_print], the path to print.
2. Add function export_dir_print, to be used to construct the path to '#{Origen.root}/vendor/lib' to print onto the exported file.
3. Modified dir: content, from dir_path to options[:dir_print]

### DIFF
--- a/lib/origen/model/exporter.rb
+++ b/lib/origen/model/exporter.rb
@@ -224,7 +224,7 @@ module Origen
         indent = ' ' * (options[:indent] || 0)
         file_path = File.join(Pathname.new(options[:file_path]).sub_ext(''), "#{id}.rb")
         dir_path = options[:dir_path]
-        dir_path_print = options[:dir_path].sub(Origen.root, 'Origen.root')
+        dir_path_print = options[:dir_path].sub(Origen.root.to_s, '#{Origen.root}')
         line = indent + "model.sub_block :#{id}, file: '#{file_path}', dir: \"#{dir_path_print}\", lazy: true"
         unless block.base_address == 0
           line << ", base_address: #{block.base_address.to_hex}"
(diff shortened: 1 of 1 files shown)


pderouen:
@williamcheong, as discussed in a separate thread you will need to fix the lint errors.

I'm a little surprised that the exporter doesn't create a relative path from origen.root by default since the intention is for the files under vendor to be checked in. @ginty any comment?

williamcheong:
@pderouen I tried to reduce the changes which get the same and expected result. Surprisingly, coverage improved, but failed continuous-integration/travis-ci/.. check.
Now the exported register file has :dir property not with hard-path but starts with Origen.root:
` dir: "#{Origen.root}/vendor/lib/models",`

But it still fail. Any idea to resolve?

pderouen:
@williamcheong the import/export spec tests are failing. You can see it in the report if you click on details and then select 1 of the builds. [Here's the report.](https://travis-ci.org/Origen-SDK/origen/jobs/475183894)

This is the first failure:

~~~
Failures:
  1) Model import and export export is alive
     Failure/Error: dir_path_print = options[:dir_path].sub(Origen.root, 'Origen.root')
     
     TypeError:
       wrong argument type Pathname (expected Regexp)
     # ./lib/origen/model/exporter.rb:227:in `sub'
~~~

pderouen:
You can run the test in your local origen area like this:

~~~
williamc/origen> ./lbin/rspec
~~~

williamcheong:
The last commit used to keep export files output directory, for version tracking of designsync.
But caused 0.008% coverage dropped and still failed integration with ruby 2.2.2.
@ginty  @pderouen  Can you help?

pderouen:
I'm not sure what to do about that failure. It's complaining that bundler isn't installed on the machine running the checks.

PR #323, opened 2019-01-15
Sub block groups

redxeth:
Added method sub_block_group to support defining an associated group of similar sub_blocks, permitting any uniqueness desired between them.

### DIFF
--- a/templates/web/guides/models/defining.md.erb
+++ b/templates/web/guides/models/defining.md.erb
@@ -188,6 +188,110 @@ $dut.nvm.reg1.address        # => 0x30
 $dut.nvm.reg1.data           # => 0xFFFF_FFFF
 ~~~
 
+#### Adding Multiple Sub Blocks
+
+##### Multi-Instance Sub Blocks
+
+If you have several similar sub_blocks, you can use the `instances:` option when defining
+a sub_block.
+
+~~~ruby
+# lib/soc/eagle_m352.rb
+def initialize(options={})
+  sub_blocks :memory,        class_name: "MEMORY_128_B954", instances: 4
+end
+
+$dut = SOC::EAGLE_M352.new
+$dut.memory0.reg1.address        # => 0x30
+$dut.memory1.reg1.address        # => 0x30
+$dut.memory2.reg1.address        # => 0x30
+$dut.memory3.reg1.address        # => 0x30
+$dut.memorys                     # => Array of all 4 memory sub_blocks
+~~~
+
+This assumes that the instances are using the same class.
+
+A different base_address can also be defined for each instance by passing an array of size equal to the number
+of instances to the base_address option:
+
+~~~ruby
+# lib/soc/eagle_m352.rb
+def initialize(options={})
+  sub_blocks :memory, class_name: "MEMORY_128_B954", instances: 4, base_address: [0x000, 0x200, 0x400, 0x800]
+end
+
+$dut = SOC::EAGLE_M352.new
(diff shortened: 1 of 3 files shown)


templates/web/guides/models/defining.md.erb
```diff
@@ -188,6 +188,70 @@ $dut.nvm.reg1.address        # => 0x30
...
+
+~~~
+
+Here's an example of a custom container class like this:
+
+~~~ruby
+# lib/memories.rb
+class MEMORIES_B294 < ::Array
+
+  def <<(sub_block)   # to meet sub_block_group requirement
```
ginty: I don't think this example should actually define the << method, by inheriting from `Array` you get that for free and don't need to define it.

I would highlight a bit more what is going in here with respect to `< ::Array`, something like:

The recommended way to create a custom container is to sub-class array and then add additional methods, like this:

~~~ruby
# lib/memories.rb
class MEMORIES_B294 < ::Array
  # Example of a custom method to print the names of all contained memories
  def print_ids
    # Since this class is based on an array, we can call use all array methods like 'each' to iterate
    # on the contained objects
    each { |mem| puts mem.id }
  end
end

sub_block_group :memories, class_name: "MEMORIES_B294"
  sub_block :memory0,        class_name: "MEMORY_128_B954"
  sub_block :memory1,        class_name: "MEMORY_64_B954"
  sub_block :memory2,        class_name: "MEMORY_32_B954"
end

# All conventional array methods are supported since we sub-classed Array
memories.size                   # => 3
memories.first                  # => <memory 0 instance>
memories.map { |mem| mem.id }   # => [:memory0, :memory1, :memory2]

# Plus we have our custom method(s)
memories.print_ids
=>
memory0
memory1
memory2
~~~

templates/web/guides/models/defining.md.erb
```diff
@@ -242,10 +305,42 @@ module NVM
...
+    def initialize(options={})
+      @size_in_kB = 32
+    end
+  end
+end
+
+# lib/memories.rb
+class MEMORIES_B294 < ::Array
+
+  def <<(sub_block)   # to meet sub_block_group requirement
```
ginty: As above, would either remove this, or change it to a different example that uses a custom container that doesn't inherit from `Array` and therefore does require this to be defined.

templates/web/guides/models/defining.md.erb
redxeth: Cool thanks.  Sure that can be a requirement.  But I think maybe added commented out `def <<(sub_block)` as an example to show in case you do want to do want to include extra stuff upon pushing?   It's what I'm doing in current project.

redxeth:
@info-rchitect No.  The container is a simple Array or a sub-class of an Array, so itself doesn't have a base_address unless you were to include that as an attribute/method of the custom container class.

templates/web/guides/models/defining.md.erb
ginty: Well in that case then I would make the example do something and also take out the comment "to meet sub_block_group requirement" which is mis-leading in this case.

templates/web/guides/models/defining.md.erb
redxeth: Yeah the reason I said that `<<` was the requirement initially is that the `sub_block_group` method explicitly uses it.   Will try to clarify that.

redxeth:
@ginty Updated the web docs.  Let me know what you think.

PR #325, opened 2019-01-16
Exporter updates

pderouen:
- Changed exported directory path to sub_block class files to use <code>"#{Origen.root!}/"</code> (relative paths)
- Added an export option <code>:rm_rb_only</code> to leave behind non-ruby code (ie revision control metadata) for cases where an existing model is being updated

@williamcheong I think this will resolve your issues with exporter

### DIFF
--- a/lib/origen/model/exporter.rb
+++ b/lib/origen/model/exporter.rb
@@ -8,8 +8,7 @@ module Origen
       #   include_registers:  true
       #   include_sub_blocks: true
       #   include_timestamp:  true
-      #   file_path:          nil     # default path is Origen.root!/vendor/lib/models/name
-      #   rm_rb_only:         nil     # delete only .rb files, default is rm -rf * Origen.root!/vendor/lib/models/name
+      #   rm_rb_only:         nil     # delete only .rb files, default is rm -rf * Origen.root/vendor/lib/models/name
       #
       # Use the rm_rb_only option if the export dir is under revision control and the dir contains revision control metadata
       def export(name, options = {})
--- a/lib/origen/model/exporter.rb
+++ b/lib/origen/model/exporter.rb
@@ -165,7 +165,7 @@ module Origen
       end
 
       def export_dir(options = {})
-        options[:dir] || File.join(Origen.root, 'vendor', 'lib', 'models')
+        options[:dir] || File.join(Origen.root!, 'vendor', 'lib', 'models')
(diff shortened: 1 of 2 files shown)


williamcheong:
@pderouen  in exporter.rb, function export_dir, need to remove the "!" of "Origen.root!". 
With "!" the actual path address was path to the gem itself, but it should be the path to the top_level destination.
The rest all works well. Great job, Thanks!

ginty:
Thanks for looking at this @pderouen.

The use of `Origen.root` vs. `Origen.root!` is confusing here for sure!

I think the change that is required by @williamcheong is that you should use `Origen.root` always in `exporter.rb`, but if the literal text `"Origen.root"` is written to the exported file, in that case it should always be `Origen.root!`.

That can be roughly summarized as - at the time of exporting you want to be referring to the current top-level app, but at the time of importing you want to be explicitly referring to the app that contains the exported file to account for the fact that it could now be running as a plugin.

pderouen:
Like that?

williamcheong:
> 
> 
> Like that?

Tested on my app, works well now. Thanks @pderouen.

ginty:
@redxeth, for something more explicit you can always do `Origen.app(:my_app_name).root`.

Otherwise open to suggestion on better aliases...

lib/origen/model/exporter.rb
```diff
@@ -1,6 +1,17 @@
...
   module Model
     module Exporter
+      # Export the model
+      #
+      # Options defaults:
+      #   include_pins:       true
+      #   include_registers:  true
+      #   include_sub_blocks: true
+      #   include_timestamp:  true
+      #   file_path:          nil     # default path is Origen.root!/vendor/lib/models/name
```
ginty: I don't think the `:file_path` option should be included here, this is not intended to be a user API - https://github.com/Origen-SDK/origen/blob/master/lib/origen/model/exporter.rb#L12

ginty: Or if you think having it there will help developers in future, please try and somehow indicate that it is a private option.

pderouen: I started out only adding documentation for the rm_rb_only option. I included file_path only because it was one of the options. I have no heartburn removing it.

pderouen: Removed it.

PR #327, opened 2019-01-30
minor stuff from BAC

info-rchitect:
- comparator fix enables relative paths
- flow change allows option to not reload target
- param change enhances fail msg
- param fix for false values

### DIFF
--- a/lib/origen/generator/flow.rb
+++ b/lib/origen/generator/flow.rb
@@ -7,7 +7,7 @@ module Origen
         # Origen.deprecated 'Origen core will no longer support program generation in v1, add the origen_testers plugin to your application to fix this warning'
         # Refresh the target to start all settings from scratch each time
         # This is an easy way to reset all registered values
-        Origen.app.reload_target! if options[:reload_target]
+        Origen.app.reload_target!
         Origen.tester.generating = :program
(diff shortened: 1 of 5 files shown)


lib/origen/generator/flow.rb
```diff
@@ -7,7 +7,7 @@ def create(options = {}, &block)
         # Origen.deprecated 'Origen core will no longer support program generation in v1, add the origen_testers plugin to your application to fix this warning'
         # Refresh the target to start all settings from scratch each time
         # This is an easy way to reset all registered values
-        Origen.app.reload_target!
+        Origen.app.reload_target! if options[:reload_target]
```
ginty: Not sure about this, seems like a breaking change.

Would prefer to see it written the other way around, so that by passing in a new option it will inhibit the reloading.

ginty: Also not sure if this actually works, does it?

This method is partially through transitioning to OrigenTesters, I think it should override what you have here - https://github.com/Origen-SDK/origen_testers/blob/master/lib/origen_testers/origen_ext/generator/flow.rb

info-rchitect: @ginty i backed out that change

Issue #329, opened 2019-01-30
How Origen handles PATH


info-rchitect:
Hi,

As I convert us to the new Origen, I see we need to set './lbin' to the user PATH.  A couple of questions:

1.  Should Origen ask users to do this or should Origen just run its commands with an augmented PATH?
2. Has anyone setup a symlink at the top of their app so users didn't have to change their PATH?  
    `bundle -> lbin/bundle`
3. How do Windows users handle the bundle executable?

thx

ginty:
This has always been part of the Origen installation instructions, the only thing that has been added here is a check/warning that a user's environment actually does it.

On Windows it should be the same, though I notice now that we never actually mentioned it in that guide.

However, as is evident from your comment, it mostly works without this. If you don't have it, it means that when you run 'origen' you are not initially launching from the Origen version that is bundled in your application. Instead it launches the Origen version that is installed to the system Ruby, but it works OK because Origen's boot system will identify this situation and silently re-launch using the bundled Origen.

Other gems won't do this though, so without 'lbin' in your PATH you could get weird errors if your application or Origen tries to launch 3rd party executables.
Potentially, you could end up running a different version of the gem than you think you are - best case that's OK because they are feature compatible, worst case it is not or else you get weird Bundler errors about different Gem versions being activated.

Examples of Origen command where this could crop up would be 'origen web compile' which calls `system('nanoc')` or `origen specs` which calls `system('rspec')`.

Unfortunately, I don't believe that there is an easy way to take care of this in Ruby code. The only thing that might partially work would be if we changed all such system calls to be `system("#Origen.root}/lbin/<executable>")`, but it would be impossible to guarantee that we got them all across plugins too.

I am open to suggestion here, is it better to allow users to be oblivious to this?

info-rchitect:
@ginty We put a symlink to lbin/bundle in our top level app, though that would not work for Windows users.  That seems to work for Linux folks.  I don't think we need to refactor a bunch, just wanted to see what folks thought.

pderouen:
I'd be in favor of removing the .\lbin from the path and updating origen to use the path from Origen.root.

Could we download all of the public apps and have a script look for "system(" and things bound by `command`?

pderouen:
...apparently \` means something in markdown.  \`some_command\` is the other way to execute a system command.

ginty:
[...]

PR #330, opened 2019-01-31
Fixed sub_block definition bug and removed lbin PATH warning

ginty:
A subtle bug (breaking change) introduced by the recent `sub_block_groups` update is that a sub_block definition stopped overriding an existing method of the same name.

This restores the original behavior but adds a warning that this is happening.
Added a similar test case for the `sub_block_groups` method.

Also made a small update to allow `sub_block[:my_block]` (singular) to work, it would have previously crashed.

Finally, removed the no lbin in PATH warning discussed in #329.


redxeth:
@ginty Thanks for this fix and adding the test case.

ginty:
@redxeth, no problem.
Man, I totally noticed that the sub_blocks_group update had added in an `unless_defined` gate around the singleton method definition, but I thought it would be OK.
Turns out it broke an app, just shows how seriously we need to be taking any change of behavior now.

spec/sub_block_spec.rb
```diff
@@ -538,6 +538,40 @@ def initialize
...
+          m = SB1.new
+          m.blah.should == 1
+          m.sub_block_group :blah do
+            m.sub_block :my_blah
+          end
+          m.blah.should_not == 1
+        end
+      end
+
+      it "adding a sub_block should not override an existing method of that name" do
```
redxeth: @ginty Should this read 'adding a sub_block should override...', i.e. remove the 'not' ?

redxeth:
@ginty generally we need to be sure to always have a test case for any expected operation like that.   Currently running into this with the 'test_ids' plugin...

info-rchitect:
@redxeth What issue with test_ids?

redxeth:
@info-rchitect let me post an issue on that github page.   We're seeing some issues with next bin from range.  I fixed a bug causing program generation to take way too long as well, but it's not yet been merged as I'm working to add a good test case.

lib/origen/sub_blocks.rb
```diff
@@ -295,10 +297,12 @@ def sub_block(name, options = {})
         if sub_blocks[name] && !sub_blocks[name].is_a?(Placeholder)
           fail "You have already defined a sub-block named #{name} within class #{self.class}"
         end
-        unless respond_to?(name)
-          define_singleton_method name do
-            get_sub_block(name)
-          end
+        if respond_to?(name)
+          callers = caller[0].split(':')
+          Origen.log.warning "The sub_block defined at #{Pathname.new(callers[0]).relative_path_from(Pathname.pwd)}:#{callers[1]} is overridding and existing method called #{name}"
```
pderouen: A small typo here. Should read, "overriding an ..."

PR #331, opened 2019-02-15
Add Pattern.create_with_sub_block_group

rlaj:
We've been doing this manually in our patterns by wrapping the Pattern.create in various kinds of loops.  This looks to clean up some of that code and let Origen natively handle iterating patterns over sub_block_groups.

~~~ruby
# Instantiate sub_block_group of multiple USB sub_block instances
sub_block_group :usbs do
  sub_block :usb0, class: 'MY_USB_CLASS'
  sub_block :usb1, class: 'MY_USB_CLASS'
end

Pattern.create_with_sub_block_group(dut.usbs) do |usb|
  # can access current instance directly
  usb.reg(:some_reg) = 0x0610_1856

end
~~~

### DIFF
--- a/lib/origen/generator/pattern.rb
+++ b/lib/origen/generator/pattern.rb
@@ -2,6 +2,7 @@ module Origen
   class Generator
     class Pattern
       include Comparator
+      include Utility::InputCapture
 
       class DummyIterator
         def invoke(*_args)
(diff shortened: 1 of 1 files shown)


redxeth:
So this will make 2 patterns for each USB or 1 for both?  If 2, then unique names are handled, otherwise the 2nd will overwrite the 1st (ask me how I know this).

rlaj:
2 separate patterns.  By default, will append sub_block name to end of pattern name.

<pattern_name>_usb0
<pattern_name>_usb1

redxeth:
Cool- thanks, I see it on line 90 now.

redxeth:
Can you override the tester.sim query if you wanted for your project ?  I may want to simulate multiple using automated regressions.

rlaj:
Was thinking about regressions as well.  How about something like  if Origen.mode == :regression?

ginty:
Thanks for the PR and I like the concept here, though I'm not sure about this initial implementation.

The official way to do this kind of thing is via the pattern iterator system (even though that is un-documented and has been superseded by source-less generation for most applications).

A couple of advantages of the iterator system is that it doesn't treat sub-blocks any differently to other properties, you can iterate on anything, and it handles target re-loading properly.
One problem with the current implementation is that the sub-block iteration is outside of the Pattern.create call (which will re-load the target each time), so the `sub_block` object passed into the pattern body is actually orphaned from the current dut.
That will probably work fine in most cases, but it would be a hell of a thing for a user to have to debug in the corner case where it doesn't.

I like the idea of Origen providing some built-in iterators though, I just think it should use the existing iterators system to implement it.

Here is an (untested) example of how to implement an iterator (at application level this would be in `config/application.rb`, within Origen you can add it to the master `Origen::Application` class):

~~~ruby
# By blocks iterator
config.pattern_iterator do |iterator|
  iterator.key = :by_blocks

  # The argument assigned to the key is passed in here and called path, this iterator should be
  # called by something like: by_blocks: 'dut.usbs'
  iterator.loop do |path, &pattern|
    # Allow anything like 'dut.usbs', 'usbs' or :usbs, but don't allow pre-fetched block(s)
    unless path.is_a?(String) || path.is_a?(Symbol)
      fail "The argument given to the Pattern.create by_blocks iterator should be a string, e.g. by_blocks: 'dut.my_blocks'"
    end
    path = path.to_s
    path = "dut.#{path}" unless path =~ /^dut/
    blocks = Array(eval(path))

    # Anything assigned to instance variables should be accessible to the setup
    # method (I think!)
    @path = path
    
    $nvm.get_blocks.each_with_index do |block, i|
      @block_index = i
      pattern.call(block)
    end
  end

  # This setup method will be called immediately before executing the pattern body and
  # gives an opportunity to do any state setup that should be done for every pattern
  # using this iterator.
  # This method can and will generate pattern content and whatever is returned by it
  # is what gets passed to the pattern body, via the |block| argument in this case.
  # In the case of iterating over rich objects, such as the block in this example it
  # performs another important function. The block object passed in here is orphaned
  # from the objects in the pattern body, this is because RGen re-instantiates the 
  # target between each iteration. This may or may not cause problems depending on the 
  # application but it could for example lead to issues if you set a register bit on the
  # orphaned object but the objects in the pattern body do not reflect that.
  # The solution to this is to use an ID or similar attribute of the block to reload
  # a new version from the current application state.
  iterator.setup do |block|
    # Reload the block object from the current environment, all top level objects 
    # accessed in here are the same ones that the pattern will use
    blocks = Array(eval(@path))
    # Pass back a valid object rather than the original orphaned version.
    blocks[@block_index]
  end

  # The default pattern name will be passed in here along with the current object,
  # whatever is returned is what will be used for the current iteration
  iterator.pattern_name do |name, block|
    name.gsub("_bx", "_block#{@block_index}")
  end
end
~~~

This would get used like this:

~~~ruby
Pattern.create by_blocks: 'dut.usbs' do |usb|
  # can access current instance directly
  usb.reg(:some_reg) = 0x0610_1856
end
~~~

Maybe you would want to support additional options to help customize how the pattern name is
modified for example.

I also haven't thought too much about how this applied to simulations...

PR #334, opened 2019-03-05
OrigenSim hooks

ginty:
This PR contains various hooks to support improvements in OrigenSim.
No updates have been made to the user-facing documentation since all of these changes are intended for internal Origen use only.

- Various log output related to pattern and program generation have been inhibited when the tester is a simulator
- The ability to abort a pattern, flow or actually any Origen job has been added by raising an `Origen::Generator::AbortError`. That will silently abort whatever is currently running and the caller is responsible for cleanup. This is used in OrigenSim to implement a max_errors attribute for simulations.
- The logger now supports interception by a 3rd party. The system that has been added is actually fairly elaborate and supports layers of interceptors where the last one added can choose whether to fully handle the log message or else process it in some way and then pass it on to the next interceptor in line. Here is an example of how to create an interceptor:

~~~ruby
# An interceptor ID is returned, this should be given to stop_intercepting
@log_intercept_id = Origen.log.start_intercepting do |msg, type, options, original|
  if some_condition_is_true?
    # Handling it ourselves
    my_method(msg, type)
  else
    # Call the original Origen.log method (or the next interceptor in line)
    original.call(msg, type, options)
  end
end

Origen.log.stop_intercepting(@log_intercept_id)
~~~

This feature is used by OrigenSim to capture all log messages during simulation so that it can synchronize them with the simulation output.

- Pins now support meta-data being supplied along with any state change:
~~~ruby
my_pin.read!(1, meta: { position: 10 })
my_pin.state_meta    # => { position: 10 }
my_pin.dont_care
my_pin.state_meta    # => {}
~~~

This API is to be used by physical protocol drivers (e.g. JTAG, SWD, SPI) to tag which pin compares correspond to register bits. This meta data is then used by OrigenSim to translate pin-level mis-compares into register actual/received data in the event of a failed read transaction.
- Testers can implement register transaction wrappers called `read_register` and `write_register` and they will be called automatically. This is used by OrigenSim to support the above.


ginty:
Also added `Origen.log.flush` to force flush any buffered log data to the console/file and duplicate patterns will not be removed from what is passed to `origen g` when the tester is the simulator.

PR #337, opened 2019-05-03
updated bundler byebug

info-rchitect:
We are moving to Redhat 7.5 and we ran into issues with the bundler and byebug gems.  Our IT will not support the older libraries needed to support byebug v8.  I hope this will work for the community.


info-rchitect:
@ginty Can you help with the build errors?  I think they are not real, rather some sort of Github configuration regarding the Bundler version.

ginty:
Thanks for the PR.

Anything in this `Gemfile` only applies when Origen core is checked out into a workspace and running from there.
Changing that is no big deal but I would prefer we change to this since it makes it clear what someone having trouble running on earlier Ruby needs to do:

~~~ruby
gem "byebug" #, "~>8"  # Keep support for Ruby 2.0
~~~

If you are having issues with the  Byebug version in an application, then likely what you really need is a similar edit within your application's Gemfile rather than Origen's.

The other change to enforce Bundler 2 is more risky. We are seeing build errors here because the CI environment does not have it available and likely this is quite representative of the kind of issues we will see if we proceed with this as it currently stands.

Why do you want to enforce Bundler 2 and have you checked that Origen works OK with it?

If you have done some verification of that then this change could be considered instead:

~~~ruby
spec.add_runtime_dependency "bundler", ">1.7", "<3"
~~~

That would open up the acceptable range to include Bundler 2, but not exclude existing application environments that run quite happily with Bundler 1.

info-rchitect:
@ginty I checked Origen 'test' and 'specs' commands and ran regression with our test interface, all passed.  I will make the changes you suggest.

priyavadan:
Restarting this discussion.

@ginty @info-rchitect I think the bundler version that will be used depends on the version that was used to bundle the Gemfile.  

Bundler 2 restricts ruby version to > 2.3 and rubygems > 2.5

If Gemfile was bundled with Bundler 1, bundler 1 will be used. If it was bundled with Bundler 2, bundler 2 will be used.

Reference: [https://bundler.io/v2.0/guides/bundler_2_upgrade.html](https://bundler.io/v2.0/guides/bundler_2_upgrade.html)

Potentially moving to RH 7.5 will expose everyone to this in my opinion. It needs to be discussed how this will be handled?

Thanks!!

ginty:
I think this change is OK, the main thing is getting the tests to run in the CI env.

The latest .travis.yml already has some changes that may help since this was created.

We could also consider dropping support for Ruby < 2.3, does anyone have any objections to that?

priyavadan:
@ginty I am okay with dropping support for Ruby < 2.3. 

The travis builds are also failing for ruby 2.4, probably because Gemfile.lock was created with Bundler 1

ginty:
We are also upgrading to RHEL7 and saw this as well today

ginty:
So the problem is not so much a RHEL6 vs 7 thing, or even a Ruby version thing.
Rather the issue is that if you ran `gem install bundler` 2 years ago, say when you were installing Ruby for your RHEL6 environment, then you would have got Bundler 1.x being installed.
However, if you run that same command today when creating a new Ruby installation then you now get Bundler 2.x installed.

If you say:

~~~
gem install bundler
gem install origen
~~~

Then your Ruby will have both Bundler 1.x and 2.x available (since Origen depends on the earlier version).

Ruby will invoke the latest version automatically when it has multiple versions available, so when `bundle` is executed within your Origen app v2 is being invoked if it is available.
This leads to an un-reconcilable conflict - Bundler v2 is already a given since it has a been invoked, yet Origen demands v1.

I was able to fix this by simply uninstalling Bundler v2 from our RHEL7 Ruby:

~~~
gem uninstall bundler --version 2.0.2
~~~

And now our apps can run happily on both RHEL6 and RHEL7.

However, this is still a problem and I shouldn't have had to do that - I think anyone installing a new Ruby and then Origen today probably can't run, which is not good!

I think the solution is just to remove the Bundler version requirement altogether and just leave it open-ended.
Since Bundler has this special place in the Ruby ecosystem, we can probably be fairly confident that even a future v3 is not going to change the public API in a way that would significantly affect us.

Taking that risk is preferable to the current situation whereby legacy apps would stop booting if their system Ruby suddenly installed Bundler v2.

ginty:
Alright, the light is green, the trap is clean.

I gave up trying to support Ruby 2.2 in the testbench, we're basically fighting against the other gems we depend on which have moved on to supporting 2.3 as the min version.

I think we just go with that, which roughly means that latest Origen will support the last 4 years of Ruby versions.
In NXP we are running 2.5, so we have ~3 years before we will need to upgrade.

priyavadan:
@ginty I am okay with this. Within AMD, I think we are using a mix of 2.3 and 2.5, so we need to make sure we can start using ruby 2.5 overall @info-rchitect

info-rchitect:
thanks @ginty !

PR #339, opened 2019-05-28
Patch for path on windows

pderouen:
Pretty minor fix to reconstruct the full path after splitting the caller on ':'

@ginty are there other places where a similar operation is done?


ginty:
Not sure about other places, possible but none that I can point to.

PR #340, opened 2019-05-29
Fix for GC bug

ginty:
Fix for an intermittent bug which we believe is caused by the paths in the `@blocks_files` being partially garbage collected in error.

This change stores the paths as strings rather than `Pathname` objects and hopefully the simpler object storage circumvents whatever the root cause is.
It seems to in repeated loop testing, I have not been able to invoke the problem with this change after running continuously for a few hours.

### DIFF
--- a/lib/origen/application.rb
+++ b/lib/origen/application.rb
@@ -50,6 +50,8 @@ module Origen
               Origen.register_application(app)
             end
             app.add_lib_to_load_path!
+            # Also blow this cache as a new app has just been added
+            @apps_by_namespace = nil
           end
         end
       end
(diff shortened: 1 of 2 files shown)


lib/origen/application.rb
```diff
@@ -50,6 +50,8 @@ def inherited(base)
               Origen.register_application(app)
             end
             app.add_lib_to_load_path!
+            # Also blow this cache as a new app has just been added
+            @apps_by_namespace = nil
```
ginty: This change has no effect on the bug in question, it would still fail with this in place. However, this could potentially be a corner case issue I think if a plugin was loaded manually by an app or something, so took the opportunity to improve it when working in this area.

PR #342, opened 2019-05-30
Include pin type in exported models

ginty:
Includes the pin type attribute in exported models.

This enables the pins file generated by the `origen sim:build` to identify WREAL pins as analog type.


ginty:
Haha, indeed @info-rchitect, one step at a time!

PR #344, opened 2019-06-11
Update to CSV data extractor to permit finding file in either app or plugin

redxeth:
Minor fix-- to let Origen file_handler help find the CSV file.

### DIFF
--- a/lib/origen/utility/csv_data.rb
+++ b/lib/origen/utility/csv_data.rb
@@ -78,7 +78,7 @@ module Origen
         result = []
         line_no = 1
         file = Origen.file_handler.clean_path_to(@file)
-        File.readlines("#{@file}").each do |line|
+        File.readlines("#{file}").each do |line|
           # Clean up line
           line.strip!
           if line =~ /^#{options[:comment_char]}/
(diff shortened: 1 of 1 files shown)


lib/origen/utility/csv_data.rb
```diff
@@ -77,7 +77,8 @@ def extract_csv_data(options = {}) # :nodoc:
         field_values = []
         result = []
         line_no = 1
-        File.readlines("#{Origen.root}/#{@file}").each do |line|
+        file = Origen.file_handler.clean_path_to(@file)
+        File.readlines("#{@file}").each do |line|
           # Clean up line
```
ginty: Doesn't look right to me, updates `file` but then goes on to reference `@file`...

redxeth: Thanks, copy typo.  Strange that it worked anyway for my 2 tests cases with the bug.  test cases : file in plugin, file in toplevel app. Fixed.

info-rchitect:
@redxeth With all of the CSV import gems are we sure we are using the best method here by rolling our own?  Maybe that is a better question outside this PR but I am asking because I get requests to import CSV all the time and have not had an answer for 'Origen recommends'.  thx

redxeth:
@info-rchitect  Yeah the CSV import capability was added a while ago-- don't have any checkin logs as it was before origen (rgen) was on github.  Anyway for now just using it and needed a little fix.

info-rchitect:
LGTM

PR #349, opened 2019-07-30
Updated web links to origen-sdk.org to use HTTPS protocol due to website

redxeth:
I haven't yet done an update for the `index.html` symlink.

Looking in the code it seems it already has code to generate a symlink for `index.html' pointing to `latest/index.html`, however it is only created if `index.html`.  

[https://github.com/Origen-SDK/origen/blob/master/lib/origen/application/deployer.rb#L155](https://github.com/Origen-SDK/origen/blob/master/lib/origen/application/deployer.rb#L155)

I think we need to update this also to delete the toplevel `index.html` if there and implement the symlink.  This should work even if symlink already there.

Thoughts?

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -40,7 +40,7 @@ class OrigenCoreApplication < Origen::Application
   
   config.web_directory = "https://github.com/Origen-SDK/Origen-SDK.github.io.git/origen"
   #config.web_directory = "git@github.com:Origen-SDK/Origen-SDK.github.io.git/origen"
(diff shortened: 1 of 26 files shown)


ginty:
Re. the index.html symlink, what's there seems OK to me.
i.e. create it if doesn't exist, but if someone has specifically created a dedicated file for it then leave it.
The site you saw with the hardcoded index.html can only be like that because someone created it manually, and in that case Origen should not blow it away. By default, it looks like Origen is already doing the right thing and would setup and maintain the site properly wrt index.html

PR #351, opened 2019-08-16
Update ruby.yml

priyavadan:
testing Github Actions


priyavadan:
@ginty interesting failure on Github actions

not sure why i have to do a bundle install for every sub-test

image

ginty:
So that is in the code that creates a graphical representation of the activity in concurrent pattern execution threads.
It looks up the width of the current console to size the graph for the current window.

I'm not sure what's more surprising, that it doesn't work here or that it worked OK in Travis!

You may need to force a hardcoded value when some environment variable is set to indicate that it is running in CI - https://github.com/Origen-SDK/origen/blob/4f0ad7b4edf181d513a1b98876b14d605c57e887/lib/origen/generator/pattern_sequence.rb#L89

priyavadan:
@ginty thanks! I'll take a look (soon)

ginty:
Actually, maybe its best to put a `begin rescue end` around that and set a default value in the event of an error.
That will mean it will naturally handle any other headless runtimes that may be used - e.g. I'm not sure if I've actually tried generating a concurrent pattern on LSF yet, maybe it will have a similar issue.

PR #354, opened 2019-08-20
fixes #353

### LINKED ISSUE #353: STDIN.noecho error
When asking the user for a session password, Origen crashes

```
Please enter your password:

COMPLETE CALL STACK
-------------------
undefined method `noecho' for #<IO:<STDIN>>
/home/priyavadan/.origen/gems/ruby/2.3.0/gems/origen-0.52.4/lib/origen/users/user.rb:200:in `password'
```

Possible solution here:  https://stackoverflow.com/questions/9324697/why-cannot-use-instance-method-noecho-of-class-io

will submit a PR with the fix.

priyavadan:
fixes #353 

I also added some code that checks if a different user_id is defined in the site config in the user directory. Running into issues where the Linux Subsystem in windows 10 uses a different user id than the corporate ID. If there is another way to change this, I am open to discussing.

### DIFF
--- a/spec/user_spec.rb
+++ b/spec/user_spec.rb
@@ -52,14 +52,4 @@ describe 'Advanced User Options' do
     Origen.current_user.id.should == original_user.id
     Origen.current_user.id.should_not == 'crradm'
   end
-
-  it 'Checks if a different user id is specified in Origen Site Config' do
-    if !Origen.site_config.change_user_id
-      Origen.site_config.change_user_id = false
-      Origen.site_config.user_id = nil
-    else
-      Origen.site_config.change_user_id = Origen.site_config.change_user_id
-      Origen.site_config.user_id = Origen.site_config.user_id
-    end
-  end
 end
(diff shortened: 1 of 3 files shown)


spec/user_spec.rb
```diff
@@ -52,4 +52,14 @@
     Origen.current_user.id.should == original_user.id
     Origen.current_user.id.should_not == 'crradm'
   end
+
+  it 'Checks if a different user id is specified in Origen Site Config' do
```
pderouen: Does this perform any tests? I only see assignment operators.

ginty:
Not sure about the use of the site-level config to do what is essentially user-level configuration.

Why not just add something like:  `User.current.override_user_id("blah")`

ginty:
I guess I can see the advantage of the site_config over a user model API because you can set it in one location to apply to all your apps, so I can see why you went that way.

I think a env variable is the way to go actually, it is already supported for the user name and email, maybe an additional one is required to override the ID - https://origen-sdk.org/origen//guides/starting/commit/?highlight=ORIGEN_EMAIL#Setting_Your_Name_and_Email

priyavadan:
@ginty updated per our discussion. I'll add some documentation updates for it later today... thanks!

PR #355, opened 2019-08-23
fixed bug when using ENV vars to set user ID

info-rchitect:


info-rchitect:
Hello, can folks please chime in here?  Would like to close this out this week.

ginty:
What's the thought behind removing the check for current user when accessing a password?

info-rchitect:
@ginty thx for the review, changes made.

info-rchitect:
`origen test -c` passes for me, seems like something is broken wrt Travis.

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated

info-rchitect:
Is there a way to manually re-run the CI jobs?

ginty:
I think this is still not quite right.

The problem is with the id method, the way it is currently implemented it will cause all user objects to pick up the ENV var, when really it should just be the user object corresponding to the current user.

I think it needs to internally keep a record of original user ID, use that to identify the current user and only then apply the ENV var. Same as how it is done for the name and email.

info-rchitect:
> if current?

@ginty How is your code possible?  I tried it and it is a chicken and egg issue in that the `current?` method is being used to assign `@id` but the return value of `current?` could possibly be `@id`.

ginty:
@info-rchitect, well I did qualify with "something along the lines of", but you will need to work out the final solution. It sounds like a change to how `@id` is assigned will also be required.

info-rchitect:
> @info-rchitect, well I did qualify with "something along the lines of", but you will need to work out the final solution. It sounds like a change to how @id is assigned will also be required.

haha so you did.  I will work on it.

PR #358, opened 2019-09-13
Fix sourceless pattern generation for sequences

ginty:
Sourceless pattern generation was broken when running `origen g` with the `--seq` option to generate a (concurrent) sequence, this fixes it.

### DIFF
--- a/lib/origen/generator/pattern_sequence.rb
+++ b/lib/origen/generator/pattern_sequence.rb
@@ -18,7 +18,8 @@ module Origen
 
       # Execute the given pattern
       def run(pattern_name)
-        ss "START OF PATTERN: #{pattern_name}"
+        name = Pathname.new(pattern_name).basename
+        ss "START OF PATTERN: #{name}"
         # Give the app a chance to handle pattern dispatch
         skip = false
         Origen.app.listeners_for(:before_pattern_lookup).each do |listener|
@@ -29,7 +30,7 @@ module Origen
           pattern = pattern[:pattern] if pattern.is_a?(Hash)
           load pattern
         end
-        ss "END OF PATTERN: #{pattern_name}"
+        ss "END OF PATTERN: #{name}"
       end
       alias_method :call, :run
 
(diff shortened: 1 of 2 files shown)


pderouen:
The changes look correct and backwards compatible to me. Do the test expectations just need update?

lib/origen/generator/pattern_sequence.rb
```diff
@@ -18,9 +18,18 @@ def initialize(name, block)
...
+        skip = false
+        Origen.app.listeners_for(:before_pattern_lookup).each do |listener|
+          skip ||= !listener.before_pattern_lookup(pattern_name.to_s)
+        end
+        unless skip
+          pattern = Origen.generator.pattern_finder.find(pattern_name.to_s, {})
+          pattern = pattern[:pattern] if pattern.is_a?(Hash)
+          load pattern
+        end
+        ss "END OF PATTERN: #{pattern_name}"
```
redxeth: Sorry if not as familiar with this code, but how does this change reconcile with similar code in generator.rb in run method when :sequence option is true?   I guess that code calls this? And since `skip` isn't passed then we need to check again?

ginty: Do you have a link? I don't see a run method in generator.rb...

redxeth: Sorry-- it was generator/job.rb.   

https://github.com/Origen-SDK/origen/blob/master/lib/origen/generator/job.rb#L154

ginty: I see.
The code in job.rb is used when looking up the top-level pattern/sequence to run. When running a pattern that is the end of it.
However, a sequence can be comprised of other patterns and sequences which are invoked via this run method.
Within the sequence you can say `seq.run :my_pattern` where the pattern name is the same as you would give to `origen g`.
The original implementation of sequences did not account for the fact that these patterns could be dispatched, so this adds the hook to handle that.

redxeth: Ok thanks-- wasn't up to speed on sequences.  I guess just make sure any other features or options used with job.rb are carried over to pattern_sequence.rb if it applies.

PR #360, opened 2019-09-23
issue 359 bug fix

info-rchitect:


info-rchitect:
We need this bug fix urgently so adding a few more folks so I can get at least 2 reviews approved before merging

info-rchitect:
> Can you change it to this which I think makes it clearer what is going on:
> self[name] ||= new_subset(name)
> Thanks

Thx @ginty thought you were off traveling, appreciate the feedback.  updates on the way.

PR #364, opened 2019-11-07
Updated clean_path_to_sub_program to support no-underscore filename

redxeth:
Added option to `clean_path_sub_program` in `file_handler.rb` to support the option of importing a sub-flow NOT pre-pended with underscore.

### DIFF
--- a/lib/origen/file_handler.rb
+++ b/lib/origen/file_handler.rb
@@ -278,10 +278,13 @@ module Origen
       path
     end
 
-    def clean_path_to_sub_program(file)
-      file = add_underscore_to(file)
+    def clean_path_to_sub_program(file, options = {})
+      options = {
+        with_underscore: true    # whether actual file in file system has underscore prepended
+      }.merge(options)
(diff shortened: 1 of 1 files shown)


redxeth:
@ginty I looked where this was used as well and only found the one spot in origen_testers.  Am thinking we move this function to OrigenTesters as part of that related PR.  It can still call the Origen FileHandler lower-level funcs, what do you think?

ginty:
Yeah I wondered about that too, ok by me

redxeth:
@ginty Ok, I've made the changes in origen_testers and created a PR https://github.com/Origen-SDK/origen_testers/pull/140.

However, not sure how to handle it here.  Would have to just add a deprecate warning I think, in case user updates origen but not origen_testers, right?

redxeth:
Added deprecate warning for those that use this version.  Don't have a version yet to put for which OrigenTesters to use, so will have to update here again once https://github.com/Origen-SDK/origen_testers/pull/140 is merged.

PR #366, opened 2019-11-20
Fix for issue #352

### LINKED ISSUE #352: parameter set API does not check for existing methods well enough
Reproduce:

1. select Ruby 2.6
2. use branch `param_set_bug_ruby_2p6`
3.  `bundle exec rspec spec/parameters_spec.rb`

This is due to the new enumerator method [chain](https://ruby-doc.org/core-2.6.3/Enumerable.html#method-i-chain). 

In reality, this is not an issue particular to the `chain` method, rather the parameter set API does not check if the set key values are available methods.  

The failing code line is [here](https://github.com/Origen-SDK/origen/blob/master/lib/origen/parameters/set.rb#L22).

I will try debugging it and make a PR but if the owners of this code could take a look it might go faster.

thx

priyavadan:
Attempting to fix #352

### DIFF
--- a/lib/origen/parameters/set.rb
+++ b/lib/origen/parameters/set.rb
@@ -88,7 +88,9 @@ module Origen
 
       OVERRIDE_METHODS.each do |method|
         define_method method do
-          method_missing(method)
+          if self[method]
+            method_missing(method)
+          end
         end
       end
 
--- a/lib/origen/parameters/set.rb
+++ b/lib/origen/parameters/set.rb
@@ -8,6 +8,7 @@ module Origen
       # Allow these parameter names to be valid. When used, they will override the
       # methods of the same name provided by the Hash class.
       OVERRIDE_HASH_METHODS = [:min, :max]
+      OVERRIDE_METHODS = [:chain]
 
       def initialize(options = {})
         if options[:top_level]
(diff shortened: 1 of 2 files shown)


ginty:
Would be good to add a test for it mind you, its easy to define param sets in the tests - https://github.com/Origen-SDK/origen/blob/master/spec/parameters_spec.rb#L385

Just create one that has a param named chain and make sure you can access it ok.

priyavadan:
will do, also think i will add some comments in the class to make it clear the need for this OVERRIDE_METHODS list.

priyavadan:
@ginty added specs, will need re-approval, thanks!

info-rchitect:
Can you make sure the chain method works at the second level as well as in the parameters set name?

priyavadan:
> Can you make sure the chain method works at the second level as well as in the parameters set name?

maybe you could write the spec test for this? ;)

just kidding, spec test on its way, should hit the parameter set name chain and a second level parameter named chain.

priyavadan:
updated spec test to actually hit the new piece of code added...

info-rchitect:
LGTM

priyavadan:
@ginty  👍or 👎

PR #367, opened 2019-11-22
enable rake file load for new application dir structure

priyavadan:


pderouen:
Are we sure there are no apps that have an app directory, but use the old structure? I don't know of any, just anticipating a possible regression.

priyavadan:
I could just skip the directory checks and add another load for app/lib/tasks/*.rake , would that be acceptable?

@ginty @pderouen

pderouen:
Sounds good to me

PR #368, opened 2019-11-25
Adding support for 'origen rc tag' without prepending 'v' to the tag

info-rchitect:
The change in the plugins file was a workaround so I could test out the PR.  **Question:** Would we ever allow someone to make a release where the session mode is set to 'debug'?

### DIFF
--- a/lib/origen/application/plugins.rb
+++ b/lib/origen/application/plugins.rb
@@ -14,17 +14,15 @@ module Origen
       # Will raise an error if any plugins are currently imported from a path reference
       # in the Gemfile
       def validate_production_status(force = false)
-        unless Origen.app.session.origen_core[:mode] == 'debug'
-          if Origen.mode.production? || force
-            if File.exist?("#{Origen.root}/Gemfile")
-              File.readlines("#{Origen.root}/Gemfile").each do |line|
-                # http://rubular.com/r/yNGDGB6M2r
-                if line =~ /^\s*gem\s+(("|')\w+("|')),.*(:path\s*=>|path:)/
-                  fail "The following gem is defined as a path in your Gemfile, but that is not allowed in production: #{Regexp.last_match[1]}"
-                end
-                if line =~ /ORIGEN PLUGIN AUTO-GENERATED/
-                  fail 'Fetched gems are currently being used in your Gemfile, but that is not allowed in production!'
-                end
+        if Origen.mode.production? || force
+          if File.exist?("#{Origen.root}/Gemfile")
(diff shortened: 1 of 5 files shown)


info-rchitect:
Works internally, not sure about at other companies

lib/origen/application/plugins.rb
```diff
@@ -14,15 +14,17 @@ def initialize
...
-        if Origen.mode.production? || force
-          if File.exist?("#{Origen.root}/Gemfile")
-            File.readlines("#{Origen.root}/Gemfile").each do |line|
-              # http://rubular.com/r/yNGDGB6M2r
-              if line =~ /^\s*gem\s+(("|')\w+("|')),.*(:path\s*=>|path:)/
-                fail "The following gem is defined as a path in your Gemfile, but that is not allowed in production: #{Regexp.last_match[1]}"
-              end
-              if line =~ /ORIGEN PLUGIN AUTO-GENERATED/
-                fail 'Fetched gems are currently being used in your Gemfile, but that is not allowed in production!'
+        unless Origen.app.session.origen_core[:mode] == 'debug'
```
ginty: Don't think we can allow this change, that will seriously change behavior for existing apps

lib/origen/version_string.rb
```diff
@@ -75,7 +75,9 @@ def valid?(_options = {})
     # Returns true if the version is a semantic format version number
     def semantic?
       !!(self =~ /^v?\d+\.\d+\.\d+$/ ||
-         self =~ /^v?\d+\.\d+\.\d+\.(dev|pre)\d+$/
+         self =~ /^v?\d+\.\d+\.\d+\.(dev|pre)\d+$/ ||
+         self =~ /^\d+\.\d+\.\d+$/ ||
```
ginty: The v is already optional above - v?, instead of just v.
I think all the change in this file are un-necessary.

info-rchitect: OK

lib/origen/application/plugins.rb
ginty: Don't like this, you can either force or you can't

ginty:
> Question: Would we ever allow someone to make a release where the session mode is set to 'debug'?

Yes, I think we do that all the time and it would be a big breaking change to enforce that

lib/origen/application/plugins.rb
info-rchitect: yeah wasn't sure, where is the force initiated?

info-rchitect:
@ginty OK I think I fixed all of your issues, pls confirm.  thx for the quick review.

PR #369, opened 2019-11-26
Fix for params chain

info-rchitect:


ginty:
What was wrong with the original version?

I don't see how the behavior of the original vs. this would be any different...

info-rchitect:
> What was wrong with the original version?
> I don't see how the behavior of the original vs. this would be any different...

Try running the last spec test with the old version

lib/origen/parameters/set.rb
```diff
@@ -61,11 +61,12 @@ def method_missing(method, *args, &block)
             self[m] = args.first
           end
         else
+          # binding.pry if $stop == 1
           if args.length != 0
             super
           else
             if !key?(method)
-              super
+              nil
```
priyavadan: @info-rchitect not sure I understand the need for this?

info-rchitect: Without that change this spec test will fail:

~~~ruby
ip.params(:not_chain).chain.should == nil
~~~

priyavadan: if I understand correctly, it'll also effect the min and max HASH methods?

ie ip.params(:it_is_a_range).max will return a nil? I could be wrong but i just want to understand

info-rchitect: let me check that test case

info-rchitect: I updated the code and the spec tests.

priyavadan: @info-rchitect thanks, it covers all the possible uses cases we have internally, looks good to me.

@ginty thoughts?

PR #373, opened 2020-01-20
removed escaping underscore in a release_note.txt

info-rchitect:
Hi,

Can someone explain the  `release_note.txt` content is applying escaped underscores during the release process?  Here is an example of the issue if the user includes URLs such as:

~~~text
# release_note.txt
* my_gem gem changed from 1.5.0 to [1.7.0](http://mywebsite.com/my_gem/latest/release_notes/#1.7.0)
~~~

The accompanying doc history is broken:

~~~text
# doc.history
* my\_gem gem changed from 1.5.0 to [1.7.0](http://mywebsite.com/my\_gem/latest/release_notes/#1.7.0)
~~~


ginty:
Yeah, this is a long standing bug, but not with a fix as simple is this.
The underscores are escaped to stop them being interpretted as markdown emphasis/bold.
The required fix is for a smarter filter which escapes the underscores unless they are within a link.

info-rchitect:
> Yeah, this is a long standing bug, but not with a fix as simple is this.
> The underscores are escaped to stop them being interpretted as markdown emphasis/bold.
> The required fix is for a smarter filter which escapes the underscores unless they are within a link.

Have any idea how I could hook this up to unit tests?

priyavadan:
@ginty agreed. Maybe a filter that checks if the _ is within a ( and )

ginty:
> Have any idea how I could hook this up to unit tests?

Should not be difficult to create a unit test - just write a method and present it with some input and verify the it returns the right thing. It doesn't need to be tested as part of the release operation.

info-rchitect:
> Have any idea how I could hook this up to unit tests?
> 
> Should not be difficult to create a unit test - just write a method and present it with some input and verify the it returns the right thing. It doesn't need to be tested as part of the release operation.

~~~ruby
  def escape_underscores(filter_links=false)
    if filter_links
      #new code
    else
      gsub('_', '\_')
    end
  end
  alias_method :escape_underscore, :escape_underscores
~~~

info-rchitect:
Think I may need help from [this gem](https://github.com/sporkmonger/addressable) so I don't make something very brittle.

info-rchitect:
What  does everyone think of this Regex for Markdown Italics?

~~~ruby
/\s*\_\S+\_\s+/
~~~

The reason I ask is because this will determine how smart this new algorithm is.  The easy way out is to use this:

~~~ruby
/\_\S+\_/
~~~

However this will insert escaped underscores in more places than perhaps the user wants.  Should we try to guess intent here?

info-rchitect:
My mother warned me about playing with regular expressions and Markdown! Taking much longer than I would like and someone will probably break it within an hr...

info-rchitect:
Ended up adding in some helpful Range and String methods I found on stackoverflow

info-rchitect:
@ginty Should the [existing doc history](https://github.com/Origen-SDK/origen/blob/master/lib/origen/application/release.rb#L326) be escaped 'smartly' as well, in case the previous bug corrupted it?

priyavadan:
> 
> 
> @ginty Should the [existing doc history](https://github.com/Origen-SDK/origen/blob/master/lib/origen/application/release.rb#L326) be escaped 'smartly' as well, in case the previous bug corrupted it?

@info-rchitect I don't think it needs to be checking existing releases, imo. @ginty thoughts?

ginty:
Yeah, agree, I wouldn't try to recover whatever is already in there.

info-rchitect:
> Yeah, agree, I wouldn't try to recover whatever is already in there.

OK then I think this is ready to merge, anyone disagree?

ginty:
> OK then I think this is ready to merge, anyone disagree?

Yeah, me. It's not actually enabled anywhere is it?

info-rchitect:
@ginty fixed

ginty:
Need to fix the build, adding a version contraint on the relevant gem in Gemfile should do it

lib/origen/core_ext/range.rb
```diff
@@ -13,4 +13,8 @@ def to_a
       a
     end
   end
+
+  def <=>(other)
+    [min, max] <=> [other.min, other.max]
+  end
```
ginty: This seems a bit heavy handed, how do ranges normally compare to each other?

priyavadan: @ginty maybe `(min..max).eql?(Range.new(other.min, other.max))` ? not sure if this has some ruby version dependencies

ginty: Dunno, but this seems to dodgy to me, overriding a primary behavior of a core Ruby class just to implement an underscore escaper.
I think this needs to be removed and add a method somewhere to do what you need.

info-rchitect: > This seems a bit heavy handed, how do ranges normally compare to each other?

Ranges do not sort based on [this discussion](https://stackoverflow.com/questions/18498738/sort-an-array-of-ranges).

priyavadan:
@ginty thanks for the quick call today! The updates to the gemspec should now make newer origen releases pass Travis CI checks for ruby 2.3.8

In addition, simplecov is also restricted to version 0.17 because 0.18 stops support for older ruby versions (<2.4)

priyavadan:
updated Gemfile.lock based on the new dependencies as well, regressions should now pass.

@ginty after confirming the pass, will merge in! Thanks for your help!

priyavadan:
@info-rchitect go ahead , merge and make the release!

PR #374, opened 2020-01-24
Report syntax errors within sub_block classes properly

ginty:
When sub-blocks are loaded by the autoloader, some errors (those related to missing references) in the sub-block files can be swallowed and in-correctly reported to the user "as unable to find (sub-block) class X".
Such cases were very hard to debug for the user!
This fixes it by being more selective about what is considered a maskable error and it now correctly reports the underlying error in such cases.

### DIFF
--- a/lib/origen/sub_blocks.rb
+++ b/lib/origen/sub_blocks.rb
@@ -478,16 +478,23 @@ module Origen
       def klass
         @klass ||= begin
           class_name = attributes.delete(:class_name)
+          tmp_class = nil
           if class_name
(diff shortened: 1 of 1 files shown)


info-rchitect:
@ginty Do you plan on adding some unit tests to raise the errors and show an example of a sub-block that now will not error?

ginty:
I wasn't planning to for this, didn't think it was worth it.
The 'happy path' of sub-blocks being resolved through this logic was/is tested though.

PR #382, opened 2020-03-06
Added new site config variable 

priyavadan:
Added new site config variable so that the sourcing of the setup file gets skipped.

The source_setup is needed for the 'older' RGen installations (which no one should be currently using anyways)

### DIFF
--- a/templates/web/guides/runtime/global.md.erb
+++ b/templates/web/guides/runtime/global.md.erb
@@ -44,7 +44,7 @@ if the site_config variable is not defined the prefix is an empty string.
 In your origen_site_config.yml file, you can specify the site_config variable as shown below
 
 ~~~ruby
-Origen.site_config.lsf_command_prefix = '/run/pkg/fs-rgen-latest/origen_setup'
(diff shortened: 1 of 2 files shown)


templates/web/guides/runtime/global.md.erb
```diff
@@ -36,4 +36,15 @@ $dut = MyApp::Falcon.new(version: 1)
...
 
+#### Origen LSF command prefix
+The Origen runtime environment can be configured to run a prefix command before running any job on the LSF.
+There is a site config variable now available to define what that prefix command is. By default,
+if the site_config variable is not defined the prefix is an empty string.
+
+In your origen_site_config.yml file, you can specify the site_config variable as shown below
+
+~~~ruby
+Origen.site_config.lsf_command_prefix = '/run/pkg/fs-rgen-latest/origen_setup'
```
ginty: I think we should lose the reference to an NXP-internal dir structure in official documentation, can you just make something up here instead?

priyavadan: @ginty absolutely, pushing update shortly!

ginty: Thanks!

PR #383, opened 2020-03-11
Add name option to Pattern.split

rlaj:
Add naming option to **Pattern.split**.  Prior to this PR, the name of the output file would have the split suffix set to "_partN".  With this update, the suffix can be defined in the pattern by passing a 'name' option to the split method.

~~~ruby
Pattern.split                 # results in pattern file <orig_pattern_name>_part1
Pattern.split(name: 'BODY')   # results in pattern file <orig_pattern_name>_BODY
Pattern.split(name: 'EXIT')   # results in pattern file <orig_pattern_name>_EXIT
~~~


rlaj:
Thanks @ginty .  Do we know why it's failing Travis for Ruby 2.3.8?  Looks like simplecov-html (~> 0.11) version, just wondering if there is already a plan to fix this.  Seems like any recent PR fails...

ginty:
Hi @rlaj, you can yank 2.3.8 from the Travis build, that's what I've been doing for other similar cases recently.
Seems like the Ruby community has moved on to 2.4 being their min version now, so we should do likewise for an easier life.

PR #385, opened 2020-03-31
Fix sub block groups

redxeth:
Minor update to upgrade how sub_block_groups are managed-- so that only a single instance is generated of a sub block group.

### DIFF
--- a/lib/origen/sub_blocks.rb
+++ b/lib/origen/sub_blocks.rb
@@ -296,7 +296,7 @@ module Origen
         Origen.deprecate 'instances: option to sub_block is deprecated, use sub_block_groups instead'
         group_name = name =~ /s$/ ? name : "#{name}s"  # take care if name already with 's' is passed
         unless respond_to?(group_name)
-          sub_block_groups group_name do
+          sub_block_group group_name do
             i.times do |j|
               o = options.dup
               o[:_instance] = j
--- a/lib/origen/sub_blocks.rb
+++ b/lib/origen/sub_blocks.rb
@@ -232,12 +232,19 @@ module Origen
     end
     alias_method :children, :sub_blocks
 
-    # Returns a hash containing all sub block gruops of given sub-block
-    def sub_block_groups(*args)
-      if args.empty?
-        @sub_block_groups ||= {}.with_indifferent_access
+    # Returns a hash containing all sub block groups thus far added
+    # if no arguments given.
+    # If given a code block, will serve as alias to sub_block_group method.
+    # Does not handle arguments, no need at this time.
+    def sub_block_groups(*args, &block)
+      if block_given?
+        sub_block_group(*args, &block)
       else
-        sub_block_groups(*args)
+        if args.empty?
+          @sub_block_groups ||= {}.with_indifferent_access
+        else
+          fail 'sub_block_groups not meant to take arguments!'
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/origen/sub_blocks.rb
```diff
@@ -232,6 +232,15 @@ def sub_blocks(*args)
...
     alias_method :children, :sub_blocks
 
+    # Returns a hash containing all sub block gruops of given sub-block
+    def sub_block_groups(*args)
+      if args.empty?
+        @sub_block_groups ||= {}.with_indifferent_access
+      else
+        sub_block_groups(*args)
+      end
+    end
```
ginty: How does this method work when called with args?

lib/origen/sub_blocks.rb
```diff
@@ -389,6 +401,10 @@ def get_sub_block(name)
       sub_blocks[name]
     end
 
+    def get_sub_block_group(name)
+      sub_block_groups[name]
+    end
```
ginty: Why is this method required?

ginty:
Thanks for the PR.
Added a couple of initial questions, and can you also provide details on what the change of behavior is?
It's not really clear from the description what problem is being solved here.
Thanks!

lib/origen/sub_blocks.rb
redxeth: Good catch-- yes this seems like a typo. 

I had to fix the `:instance` option case in the `sub_block` method to call `sub_block_group` (singular) as it failed a test case (line 299), so I assumed that this would check this as well but now it seems recursive in nature.  Maybe we don't have a test case? 

Let me run it some more with correction.

lib/origen/sub_blocks.rb
redxeth: This is what the singleton method for sub block groups calls. Just doing similar to what is done for sub blocks. I guess I could have put it in the singleton method directly -- just was using the same structure as for sub blocks.

ginty: Ok.
Generally I don't like methods called `get_` as part of a user API, not very Rubyish, and in this case even less when its a straight wrapper around an existing accessor, `sub_block_groups`.
Also looking above, I can see this method being called on line 377, and then the unwrapped version being used on line 389 (which is fine), so overall I think we should not be adding this redundant method here.

redxeth:
Basically, (as I said in our call), previously the singleton method defined to support the sub block group would create a brand new instance of the sub block group class each time you called it (and thus called the singleton method). 

See https://github.com/Origen-SDK/origen/blob/master/lib/origen/sub_blocks.rb#L366-L370.   The `Object...new` here would create a new instance.

What this meant was that if you used a class to define your sub block group container (which I did) and wanted to have instance variables then those variables would get lost / reset every time you referenced the sub block group.

e.g.

~~~ruby
> dut.my_sub_block_group_name.value_to_remember = 5
> dut.my_sub_block_gorup_name.value_to_remember
  nil
~~~

Now the sub block group container is instantiated once, then that instance retrieved via the singleton method each time it is called, similar to how sub blocks themselves are handled.

~~~ruby
> dut.my_sub_block_group_name.value_to_remember = 5
> dut.my_sub_block_gorup_name.value_to_remember
  5
~~~

lib/origen/sub_blocks.rb
redxeth: Was just emulating what was there already-- I can clean them both up if you like.

Issue #387, opened 2020-05-05
Coverage appears broken for the new app structure


priyavadan:
I believe that "origen specs -c" is broken for apps that use the new application structure (app/lib instead of lib/)

The code that needs to be updated is https://github.com/Origen-SDK/origen/blob/master/lib/origen/commands.rb#L66 

Also need to add another directory location here https://github.com/Origen-SDK/origen/blob/master/lib/origen/commands.rb#L76

I will submit a PR for this issue.

priyavadan:
looks like I will need some assistance. I made the changes I thought would be okay for these but [this](https://github.com/Origen-SDK/origen/blob/master/lib/origen/commands.rb#L189) line fails and I cant seem to understand why. I do feel that this is a SimpleCov configuration problem related to grouping the different files.

priyavadan:
never mind, I think I figured it out

PR #388, opened 2020-05-06
Bugfix/coverage check

priyavadan:
Attempt to fix issue #387

### DIFF
--- a/lib/origen/commands.rb
+++ b/lib/origen/commands.rb
@@ -62,25 +62,34 @@ if ARGV.delete('--coverage') ||
   Origen.log.info 'Started code coverage'
   SimpleCov.configure do
     filters.clear # This will remove the :root_filter that comes via simplecov's defaults
+
     add_filter do |src|
-      !(src.filename =~ /^#{Origen.root}\/lib/)
-    end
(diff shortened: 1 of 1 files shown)


lib/origen/commands.rb
```diff
@@ -62,24 +62,31 @@
...
 
     # Results from commands run in succession will be merged by default
     use_merging(!ARGV.delete('--no_merge'))
-
     # Try and make a guess about which directory contains the bulk of the application's code
     # and create groups to match the main folders
     d1 = "#{Origen.root}/lib/#{Origen.app.name.to_s.underscore}"
     d2 = "#{Origen.root}/lib/#{Origen.app.namespace.to_s.underscore}"
     d3 = "#{Origen.root}/lib"
+    d4 = "#{Origen.root}/app/"
```
ginty, line 81: I think we should make this highest prio, for an app under the new structure the app dir should contain the majority of their code even if they happen to have an old style lib/ dir for some reason

priyavadan: sounds good!

priyavadan: @ginty pushing updates now

PR #390, opened 2020-05-24
Feature/regression and log

CodyBriscoe:
- added on on_interface_reset check at the end of reset_interface, since I needed to adjust Origen.interface.flow.subdirectory immediately after the interface was created

- relog can now relog to a custom log, via options[:log_file]

- custom log methods can now output to console if passed verbose: true

- workspace manager can now pass a build_method option, if :clone, then it won't try to init the workspace or make the empty folder

- regression_manager can utilize a custom log

- regression_manager won't delete lbin if Origen is using version 0.40.0 or higher

- fixed typo in submit_origen_job

### DIFF
--- a/lib/origen.rb
+++ b/lib/origen.rb
@@ -610,6 +610,9 @@ unless defined? RGen::ORIGENTRANSITION
         else
           @interface.reset_globals if @interface.respond_to?(:reset_globals)
         end
+        if @interface.respond_to?(:on_interface_reset)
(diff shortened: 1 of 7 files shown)


priyavadan:
@CodyBriscoe I'll merge this tomorrow and make a new release. Great work and thanks for the PR!

priyavadan:
Making release now...

PR #393, opened 2020-06-24
Add absolute value checks to as_units for negative inputs

JamesCreamer:
This is to handle negative as_units inputs the same as positive inputs. If we do not compare absolute value, we will fall through to the else block for all negatives.

Current:
irb(main):001:0> 10.as_units('[mA]')
=> "10[mA]"
irb(main):002:0> -10.as_units('[mA]')
=> "-1.000e+01[mA]"
irb(main):003:0>

Desired:
irb(main):001:0> 10.as_units('[mA]')
=> "10[mA]"
irb(main):002:0> -10.as_units('[mA]')
=> "-10[mA]"

### DIFF
--- a/lib/origen/core_ext/numeric.rb
+++ b/lib/origen/core_ext/numeric.rb
@@ -54,27 +54,27 @@ class Numeric
   end
 
   def as_units(units)
-    if self.abs >= 1_000_000_000_000_000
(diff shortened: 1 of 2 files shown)


priyavadan:
@JamesCreamer thanks for the PR and Welcome! This looks good to me.

@ginty thoughts?

PR #398, opened 2020-08-21
disable lambda style to prevent lint block

CodyBriscoe:
With this enabled, there's a scenario where its impossible to get lint to pass, for example in a v93k flow with a multiline proc in the on_fail option:

``` ruby
func :my_test, on_fail: -> do
   # on fail tests
end
```
results in
`Style/Lambda: Use the lambda method for multi-line lambdas.`

So, moving to lambda style:
``` ruby
func :my_test, on_fail: lambda do
   # on fail tests
end
```
throws up a stack trace when I run origen program:
```
COMPLETE CALL STACK
-------------------
tried to create Proc object without a block
```
Found out that to get that to work, I have to use braces {}:
``` ruby
func :my_test, on_fail: lambda {
   # on fail tests
}
```
Which now generates correctly with origen p, but now I'm back to a lint fail:
`Style/BlockDelimiters: Avoid using {...} for multi-line blocks.`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: ruby
 rvm:
-  - 2.4.7
   - 2.5.6
   - 2.6.4
 before_install:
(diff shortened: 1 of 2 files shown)


priyavadan:
@ginty  we might need to lock simplecov / simplecov-html to 0.17.0 and 0.10.0 respectively if we still want to check ruby 2.4.x on travis CI.

ginty (approved):
This looks good to me, I would also support actually implemeting the opposite rule (require lambda to change to ->) if there is an available rule for that.

I'm also OK with dropping support for 2.4.

Thanks for the update @CodyBriscoe!

PR #400, opened 2020-09-02
added capability to update_packages after a pin has been defined

priyavadan:


info-rchitect:
I like this PR but have one question:  Can the user also pass in package specific metadata in this API?  Or will it be a 2-step process?

priyavadan:
I guess after a package has been added to a pin, you could use add_location and add_dib_meta to update information for that particular package?

either ways, I want to close this PR and we can open another one if another update is needed.

priyavadan:
> @priyavadan, I made this mistake too when writing that comment, the one in the middle is real code

@ginty understand, I think the latest updates should be correct!

PR #403, opened 2020-09-21
rspec 3 formatter support

CodyBriscoe:
When using rspec 3.x, a failing spec test would result in a stack trace rather than rspecs helpful summary output of failures.

Now support for rspec3, while maintaining < rspec 3 versions


priyavadan (approved):
This looks great!

ginty (approved):
LGTM, thanks!

PR #404, opened 2020-10-01
don't assume how many site configs are found in origen

CodyBriscoe:
I tried setting up an ORIGEN_CENTRALIZED_SITE_CONFIG environment variable and ran into some issues.

When I point my app to a clone of the origen repo, I find 2 site configs as origen expects:
```
0: /.../origen/origen_site_config.yml (local)
1: /.../origen/origen_site_config.yml.erb (local)
```

When I just use a gem dependency however, I just find 1:
```
0: /.../.origen/gems/ruby/2.6.0/gems/origen-0.57.2/origen_site_config.yml (local)
```

The issue then is that setting a ORIGEN_CENTRALIZED_SITE_CONFIG variable, origen assumes its going to find 2 site configs, and hardcodes its insertion into the configs array with an index of -3, which then fails out:
```
/.../ruby-2.6.0/lib/ruby/gems/2.6.0/gems/origen-0.55.3/lib/origen/site_config.rb:421:in `insert': index -3 too small for array; minimum: -2 (IndexError)
```

Note, I'm still working out how to validate this properly, our environment makes this particular fix tricky to trial out

### DIFF
--- a/lib/origen/site_config.rb
+++ b/lib/origen/site_config.rb
@@ -416,9 +416,10 @@ module Origen
       # Load any centralized site configs now.
       centralized_site_config = find_val('centralized_site_config')
       if centralized_site_config
-        # We know the last two site configs will exists (they are in Origen core) and that they contain the default
-        # values. We want the centralized config to load right after those.
-        @configs.insert(-3, Config.new(path: centralized_site_config, parent: self))
+        # The first site configs found will exist in Origen core, and they contain the default values.
+        # We want the centralized config to load right after those.
+        centralized_index = -@configs.select { |c| c.path.start_with?(Origen.top.to_s) }.size
+        @configs.insert(centralized_index, Config.new(path: centralized_site_config, parent: self))
       end
 
       # After all configs have been populated, see if the centralized needs refreshing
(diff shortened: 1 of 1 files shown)


info-rchitect (approved):
LGTM, if you find a way to test this great, if not may need someone at another company to run it.

coreyeng:
Are you guys placing additional site configs (or removing them, I guess), inside of your Origen install itself?  I don't see how those two shouldn't be the *only* two unless `Origen.top` is being further configured internally.

priyavadan:
@coreyeng our Origen install is a 'gem' and not a package (like at NXP) so it comes with the ruby version installed on the systems. As a result running 'origen site configs' only displays the yml file and not the yml.erb

~~~
The following config files are ordered from last-encountered (highest priority) first-encountered (lowest priority)

0: /path/to/package/repo/ruby-2.6.0/lib/ruby/gems/2.6.0/gems/origen-0.59.3/origen_site_config.yml (local)

~~~

priyavadan:
@coreyeng any concerns with merging this ?

coreyeng (approved):
Ah, sorry! Forgot about this. Yes, this is fine. I was more curious as to why it was needed, but your comment makes sense.

priyavadan:
@CodyBriscoe is this ready to merge?

PR #407, opened 2020-10-30
Updated pin type to be set automatically unless set by user

info-rchitect:
The reason for this change is so we can put the pin type in the metadata automatically vs having to set it.  It is backwards compatible.

### DIFF
--- a/spec/pins_v3_spec.rb
+++ b/spec/pins_v3_spec.rb
@@ -70,6 +70,7 @@ describe "Origen Pin API v3" do
       $dut.pin[:pinz].direction.should == :io
       $dut.pin[:pinz].rtl_name.should == "pinz"
       $dut.pin[:piny].rtl_name.should == "piny[0]"
+      $dut.pin[:piny].type.should == :signal
     end
 
     it "pins know what packages they are in" do
@@ -745,6 +746,7 @@ describe "Origen Pin API v3" do
       $dut.power_pins(:vdd0).voltage = [1.2,2.5]
       $dut.power_pins(:vdd0).voltage.should == [1.2,2.5]
       $dut.power_pins(:vdd0).voltages.should == [1.2,2.5]
+      $dut.power_pins(:vdd0).type.should == :power
     end
 
     it "power pin data can be set at instantiation time" do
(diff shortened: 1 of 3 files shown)


priyavadan (approved):
LGTM Thanks!

info-rchitect:
Any issues here preventing merge?

pderouen:
@info-rchitect I haven't responded because I haven't ever paid attention to the pin type and don't know where that information gets consumed. The only thing that raises a question to my mind is that it now defaults to a pin type that is new in this PR. As long as that doesn't cause backward compatibility problems, I don't have any issue with merging.

spec/pins_v3_spec.rb
```diff
@@ -70,6 +70,7 @@ def add_packages
       $dut.pin[:pinz].direction.should == :io
       $dut.pin[:pinz].rtl_name.should == "pinz"
       $dut.pin[:piny].rtl_name.should == "piny[0]"
+      $dut.pin[:piny].type.should == :signal
```
pderouen, line 73: Was this nil or digital before?

info-rchitect:
@pderouen  Thanks for the feedback.  You do bring up a corner case where some apps at various companies could be expecting a pin type to be nil.  I don't think that would be a common case but it is a real possibility.  The PR is backwards compatible with all of the origen spec tests.

spec/pins_v3_spec.rb
info-rchitect: nil, `:analog` and `:digital` were the options for setting the pin type previously

pderouen:
It seems reasonable to have a default type of signal. I guess if the tests pass and there are no objections it should be fine.

info-rchitect:
> It seems reasonable to have a default type of signal. I guess if the tests pass and there are no objections it should be fine.

Not all pins default to `:signal` just to be clear.  If it is a ground pin it will default to `:ground`, etc.

pderouen (approved):
Seems like a good change to me

priyavadan:
@chrishume @coreyeng @chrisnappi can ya'll confirm this change would be okay within NXP? Since all spec tests are passing currently, I don't expect there to be issues but just want to confirm.

PR #409, opened 2020-11-11
Fixed tester.handshake crash for concurrent pats

ginty:
This stops Origen crashing when calling tester.handshake for the V93K in a concurrent pattern scenario.

90% of the diff here is adding a new test case, the fix is just a simple bandaid added to `lib/origen/generator/pattern_sequence.rb`


pderouen (approved):
It's better than the alternative.

ginty:
Included in Origen 0.59.6

Issue #415, opened 2020-12-08
Origen not driving pins at time 0 in simulation


jiangliu23:
When adding a new pin and initializing the state of the pin using code like below, it appears that the pins are not being driven to the expected state in simulation at T0, the versions of Origen that have broken this seems to be 0.59.4 and 0.59.5

~~~ruby
add_pin :io_1, direction: :input, reset: :drive_lo
~~~

So Origen 0.59.3 (passes with no issue)
Origen 0.59.4 (broken)
Origen 0.59.5 (broken)

ginty:
Pretty sure this is related to https://github.com/Origen-SDK/origen/pull/407

@info-rchitect, @priyavadan, I think we need to consider rolling back on that.

I'm not sure exactly where the problem lies, but I would guess somewhere it is assuming that a nil pin type means digital (vs. analog day) and now that code is broken because such pins now have a pin type called signal.
Something along those lines I think.

info-rchitect:
@ginty I would just roll it back until we can figure out where the gap in the spec tests exist that would cover this unknown connection.  Let me see when I can get to it.

ginty:
I think the problem in this case is caused by this method - https://github.com/Origen-SDK/origen_sim/blob/master/lib/origen_sim/origen/top_level.rb#L8

Previously if you called it with `rtl_pins(type: :digital)` then it would also return pins with no type, but since such pins now have a type of `:signal` they now don't get returned.

We could patch that method easily enough, however I think pins going from having no type by default to having a type is a change in behavior which we should not do - its possible other apps may run into issues when they upgrade in the future and would rightly call this a potentially breaking change.

So what do you want to do, roll back on that update completely? Or keep most of it around and just remove the defaulting to `:signal`?
The latter would allow you to either assign `:signal` as the type within your app, or else implement a default of `nil == :signal` within your app.

info-rchitect:
@ginty https://github.com/Origen-SDK/origen/pull/416

I would just revert and force people to set it when adding pins.  I do think we should add a spec test to protect the current behavior.  I will make a new PR that allows people to set the pin type to something other than `:digital` or ':analog`.  Let me know if you agree.

ginty:
Ok thanks @info-rchitect, sounds good!

jiangliu23:
> @ginty #416
> 
[...]

PR #416, opened 2020-12-08
reverted automatically setting pin type

info-rchitect:
Related to issue #415 

@ginty If you point me in the right direction i can try to add a spec test for the current behavior that got broken.

### DIFF
--- a/lib/origen/pins/pin.rb
+++ b/lib/origen/pins/pin.rb
@@ -23,7 +23,7 @@ module Origen
       PACKAGE_SCOPED_ATTRIBUTES = [:location, :dib_assignment, :dib_meta]
 
       # Pin Types, 'digital' and 'analog' are legacy types kept for backwards compatibility
-      TYPES = [:analog, :digital]
+      TYPES = [:analog, :digital, :signal, :ground, :power, :virtual]
 
       attr_accessor :order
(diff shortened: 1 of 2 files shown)


ginty:
Thanks a lot for the quick patch!

I think the test just needs to be `pin(:blah).type.should == nil` whenever no type is specified.

As long as that passes, you can keep around your additional types if you want?

info-rchitect:
> Thanks a lot for the quick patch!
> 
> I think the test just needs to be `pin(:blah).type.should == nil` whenever no type is specified.
> 
> As long as that passes, you can keep around your additional types if you want?

OK will keep the additional types, one sec.

priyavadan:
@info-rchitect looks good to me, thanks for the quick fix!

info-rchitect:
Good to merge?  If so I will make a patch release.

priyavadan:
ok from my end. 

@Jiangliu23 can you confirm if this branch fixes your use case/issue ?

jiangliu23:
> ok from my end.
> 
> @jiangliu23 can you confirm if this branch fixes your use case/issue ?

thanks for the fix, i just confirmed in my environment, this branch no longer shows the issue. cheers.

priyavadan (approved):
good to go!

PR #417, opened 2020-12-17
Fix loader bug that manifested in CrossOrigen

ginty:
This fixes a bug that showed up in the CrossOrigen test cases.

It fixes that but unfortunately I've not been able to reproduce it in a test case within Origen core.

The intention of the changed line is to check if a const reference like `MY_CLASS` exists at the top-level namespace. However, under certain conditions (could be a Ruby 2.5 vs 2.6 thing for example) this coupled to the calling code such that something like `Placeholder::BLAH:BLAH::MY_CLASS` is what actually got checked.

The addition of the leading `::` enforces the original intent and all existing specs pass plus the specs in CrossOrigen now pass too.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,8 +1,8 @@
 GIT
   remote: https://github.com/Origen-SDK/origen_app_generators.git
-  revision: 5e2dc47fd20a8b2c625e5044def2c66804ca95b0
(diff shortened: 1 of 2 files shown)


pderouen (approved):
Fixes the cross_origen crash I was seeing

chrishume:
Changes look good, but I may not be able to try them out for a few days (on PTO).

chrishume (approved):

PR #421, opened 2021-01-11
updated instructions for windows install

pderouen:
Do these updated instructions look good? @coreyeng I left the rubygem version update in the instructions. Have you encountered any problems with the version that is pacakged with ruby 2.7 (3.1.4)?


pderouen:
Is anybody looking into the problem with the regression tests?

info-rchitect (commented):
Is 2.7 required or just your selection?

pderouen:
@info-rchitect it is just my selection.

pderouen:
...may need to change the Ruby version to 2.6 as spec tests fail with 2.7 (unless we use a newer version of the dentaku gem: "~>3")

pderouen:
> 
> 
> ...may need to change the Ruby version to 2.6 as spec tests fail with 2.7 (unless we use a newer version of the dentaku gem: "~>3")

I can either add the origen.gemspec update to this PR or downgrade the recommended Ruby version to 2.6. I'll defer that decision to somebody else.

info-rchitect:
@pderouen I would change to 2.6 as we had issues internally with 2.7 as well

pderouen:
> 
> 
> @pderouen I would change to 2.6 as we had issues internally with 2.7 as well

I will change it. Choco doesn't seem to offer a ruby 2.6 package. I'll just change the instructions to "install ruby" and provide a link to the ruby website with a recommendation to install 2.6.

The only issue with 2.7 I encountered was the 1 old gem version that used a deprecated interface. Changing that 1 dependency allowed all of the regression checks to pass on Windows with ruby 2.7. Just FYI, this is sort of a separate issue.

priyavadan:
Just so you know I did try with ruby 2.7 in the past and the pin v3 spec tests fail , so if someone has the time and is looking into upgrading it , it would be worthwhile to run the spec tests too

pderouen:
I wasn't really looking into upgrading :) But, the only change I had to make in order for a tests to pass with 2.7 was to update the version of dentaku

info-rchitect (approved):
LGTM (without actually trying the instructions)

pderouen:
> 
> 
> LGTM (without actually trying the instructions)

Awesome! Thanks. Feel free to push any updates to this branch. I'm getting Ruby 2.6 installed now.

I feel like there's a potential for non-Windows savy folks to get the step updating the path wrong. If you have any suggestions there let me know.

pderouen:
After the meeting last Thursday, we decided to simplify further and move away from using choco.

@priyavadan, @coreyeng how do these updates look good to you?

priyavadan (approved):
LGTM!

PR #422, opened 2021-01-20
Fix Github Actions

pderouen:
Disabled the web compile check for Windows. On GA the gems are installed in C: and the repo is cloned and executed in D:

This causes run time errors during web compile when template files from plugins located in C: are compiled and the <code>relative_path_to</code> function is called. Some of the code that would need to be fixed is in the doc helpers plugin. I don't know that it's worth the effort to fix it. I'd prefer to have GA clone to C: instead of D: (if that's a thing).

I consider this a stop gap solution to get the tests working again.


priyavadan (approved):
Thanks for looking into this @pderouen . I am okay with this stop-gap solution, specially since the linux world does have the web compile check happening.

PR #423, opened 2021-01-20
Add support for Ruby 2.7

pderouen:


origen.gemspec
```diff
@@ -48,7 +48,7 @@ Gem::Specification.new do |spec|
   spec.add_runtime_dependency "scrub_rb", "~>1.0"
   spec.add_runtime_dependency "gems", "~>0.8"
   spec.add_runtime_dependency "highline", "~>1.7"
-  spec.add_runtime_dependency "dentaku", "~>2"
+  spec.add_runtime_dependency "dentaku", "~>3"
```
pderouen, line 51: can/should this dependency only be updated if running some minimum ruby version? The issue with 2.7 was syntax that was deprecated several versions ago that has finally been removed in 2.7. So, this dependency change will break some older Ruby version.

pderouen: In the call this week we decided that as long as the tests work back to 2.5, this is fine

pderouen:
Want to take this change?

info-rchitect (approved):
LGTM

PR #424, opened 2021-01-21
Win ga web compile

pderouen:
Fix for GA Windows Web compile check. The call to relative path generates a run time error when the drives are mismatched. There is a required bug fix to doc helpers to enable this check for Windows. Gemfile will be reverted once that patch is released.

See https://github.com/Origen-SDK/origen_doc_helpers/pull/16


pderouen:
That should be it to have the web compile check working for windows with github actions.

pderouen:
Any thoughts on this? Pretty simple change.

priyavadan (approved):
LGTM!

PR #426, opened 2021-04-15
support for remote block inheritance and overriding existing blocks

CodyBriscoe:
This is to support the direction we're heading in for supporting a chiplet based distributed architecture. 

The application structure consists of 3 layers: IP -> Chiplet -> Product

So say for example theres a PCIe App, an I/O App, and an app for Product A.

The PCIe app instantiates a pcie block and derivative: app/blocks/pcie/derivatives/pcie4

The I/O app instantiates an iod that uses pcie4: app/blocks/iod/derivatives/my_iod
```
# app/blocks/iod/derivatives/my_iod/sub_blocks.rb
sub_block :pcie, class_name: 'PcieApp::Pcie::Pcie4'
```
And then the product app instantiates an iod sub block:
```
# app/blocks/dut/derivatives/product_a/sub_blocks.rb
sub_block :iod, class_name: 'IodApp::Iod::MyIod'
```
-----
Up to this point, if I understand the distributed architecture correctly, then nothing out of the norm at this point.

However, we need to be able to support product level and chiplet level customizations. Now we could try to just pass options upstream, but a requirement at our company is that the product app has final say on the contents of the test program. 

This PR enhances the sub_block API with the `inherit` and `override` functionality.
`inherit` takes an argument similar to `class_name` but of another block's class_name. It essentially just ensures that the block loader includes the block files of the inherited block. For example, if the IO App wanted to create its own pcie sub block to customize the pcie4 block from the pcie app:
```
# app/blocks/iod/derivatives/my_iod/sub_blocks.rb
sub_block :pcie, class_name: 'IodApp::Iod::MyIod::Pcie', inherit: 'PcieApp::Pcie::Pcie4'
```
Now we can build up chiplets to be re-used across products and customize them as needed.

But now we want to customize the dut.iod.pcie block within the product app, except now that block already exists. To do that:
```
# app/blocks/dut/derivatives/product_a/sub_blocks/iod/sub_blocks.rb
sub_block :pcie, class_name: 'ProductApp::DUT::ProductA::Iod::Pcie', inherit: 'IodApp::Iod::MyIod::Pcie', override: true
```
-----
Leaving this as a draft so this can be reviewed while I add unit tests and documentation

### DIFF
--- a/lib/origen/sub_blocks.rb
+++ b/lib/origen/sub_blocks.rb
@@ -317,10 +317,29 @@ module Origen
         # been instantiated yet. This is not supported yet for instantiated sub-blocks since
         # there are probably a lot more corner-cases to consider, and hopefully no one will
         # really need this anyway.
-        if sub_blocks[name] && !sub_blocks[name].is_a?(Placeholder)
-          fail "You have already defined a sub-block named #{name} within class #{self.class}"
+        # Note that override is to recreate an existing sub-block, not adding additional
+        # attributes to an existing one
+        if options[:override]
+          sub_blocks.delete(name)
+          # this is to handle the case where a previously instantiated subblock wont allow
+          # the current class name to exist
+          # e.g. NamespaceA:B:C
+          # =>  NameSpaceX:Y:Z
+          # After requiring the files, constants become sane again:
+          # e.g. NamespaceA:B:C
+          # =>  NameSpaceA:B:C
+          if options[:class_name] != options[:class_name].constantize.to_s
+            block_dir = options[:block_file] || _find_block_dir(options)
+            Dir.glob("#{block_dir}/*.rb").each do |file|
+              require file
+            end
+          end
+        else
+          if sub_blocks[name] && !sub_blocks[name].is_a?(Placeholder)
+            fail "You have already defined a sub-block named #{name} within class #{self.class}"
+          end
         end
-        if respond_to?(name)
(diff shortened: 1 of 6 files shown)


lib/origen/sub_blocks.rb
```diff
@@ -404,6 +423,33 @@ def namespace
...
+          return _find_block_dir(options, current_path.join("sub_blocks/#{current_namespace}"), remaining_namespace)
+        else
+          Origen.log.error "Could not find block dir for namespace #{options[:class_name]}!"
+          fail
+        end
+      else
+        if current_path.join('model.rb').exist?
+          return current_path.to_s
+        else
+          Origen.log.error "Could not find block dir for namespace #{options[:class_name]}!"
```
priyavadan, line 447: should this say could not find model file for namespace?

CodyBriscoe: I had this still stay couldn't find block dir, since it tried to search down to the path that matches the block's namespace and even though the actual issue is it didn't find the model.rb, I'm assuming (maybe incorrectly) that this means it wasn't able to find the correct folder containing the class definition files

priyavadan:
lgtm!

ginty (approved):
This LGTM too, seems well tested and documented!

info-rchitect (approved):
LGTM, will be nice to see it in action

PR #427, opened 2021-06-16
skip requiring non essential block files, updated docs for sub block override

CodyBriscoe:
the requires that were helping make the constants sane starting getting troublesome when it tried to require say an attributes.rb file that called code that only worked after the blocks above it were set

Now only require files that should develop a namespace (ie model.rb/controller.rb by default), also allow for user to customize if they need to.

See the guides update for more detail


CodyBriscoe:
Converted to draft, ran into some issues and I may need to tinker some more

CodyBriscoe:
The issues turned out to be unrelated, PR ready for review again

priyavadan (approved):
LGTM! Thanks!

PR #430, opened 2021-08-10
Feature/invocation

priyavadan:
use site config to enable/disable origen invocation tracking as well as set the invocation  url

### DIFF
--- a/lib/origen/application/runner.rb
+++ b/lib/origen/application/runner.rb
@@ -151,7 +151,7 @@ module Origen
            rescue
             # Don't allow this to kill an origen command
           end
-        yield
+          yield
           begin
             unless Origen.running_remotely?
               # Wait for a server response, ideally would like to not wait here, but it seems if not
--- a/lib/origen/application/runner.rb
+++ b/lib/origen/application/runner.rb
@@ -134,30 +134,36 @@ module Origen
       #
       # @api private
       def record_invocation(options)
-        # record_invocation = false
-        # begin
-        #  # Only record user invocations at this time, also bypass windows since it seems
-        #  # that threads can't be trusted not to block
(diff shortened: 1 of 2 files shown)


priyavadan:
tests are passing in the local environment, not sure what is wrong with Github Actions

image

pderouen:
https://github.com/Origen-SDK/origen/blob/b232538178b80e566e78d4e74d8c5c6807b800ec/spec/callbacks_spec.rb#L148-L150

Looks like that failing test has a history of behaving differently between local and CI runs.

pderouen:
It fails locally for me the same as with CI

image

priyavadan:
Interesting, none of the changes affect callback testing though.

Wonder why it fails locally for you but not for me...

pderouen:
I'm poking around to see if I can figure out why. But, <code>program_generated</code> is not getting called.

lib/origen/application/runner.rb
```diff
@@ -134,30 +134,40 @@ def prepare_and_validate_workspace(options = {})
...
-        #  # that threads can't be trusted not to block
-        #  unless Origen.running_remotely? # || Origen.running_on_windows?
-        #    record_invocation = Thread.new do
-        #      Origen.client.record_invocation(options[:action]) if options[:action]
-        #    end
-        #  end
-        # rescue
-        #  # Don't allow this to kill an origen command
-        # end
+        if Origen.site_config.record_invocation == true
```
ginty, line 137: I assume this just returns `nil` today for any attribute not mentioned in the config?

priyavadan: correct!

ginty (approved):
LGTM

pderouen (approved):

PR #431, opened 2021-08-10
comment out default values for the LSF Grid from the origen_site_conf…

priyavadan:
…ig.yml file

### DIFF
--- a/origen_site_config.yml
+++ b/origen_site_config.yml
@@ -146,9 +146,11 @@ gem_use_from_system:
 #ldap_user_id_attribute: uid
 
 # LSF Configuration
-lsf_queue: 'batchq'
-lsf_group: 'nil'
-lsf_project: 'msg.te'
-lsf_resource: 'rhel6'
-lsf_cores: '1'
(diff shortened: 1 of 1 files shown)


pderouen:
I have no idea what the implications of this change are. But, it seems fine to me. Why are the GitHub Actions not running?

priyavadan:
good question, i do not know why Github Actions did not  run, maybe because the code itself  did not change just a configuration file that isnt used in the spec tests?

based on my conversation with @ginty , these changes should be okay for NXP's use case too @pderouen

coreyeng:
Yes, these should be fine for us. I know 2 of them are even correct anymore and 1 I'm pretty sure isn't, but not 100%. The other is nil and number of cores we usually boost anyway. We have our own site config I can request changes to if need be as well, and it'd only effect Windows users who update or when we invoke globally from this release forward.

TL;DR, its fine.

coreyeng (approved):

priyavadan:
@pderouen  any concerns?

pderouen (approved):
No concerns

PR #434, opened 2022-01-22
block inheritance feature can now propagate bugs and features

CodyBriscoe:
When using the sub block inheritance feature, bugs and features can now propagate. This enables us along with the existing remote inheritance functionality to split bug/feature definitions into apps that are targeted towards modeling some IP, while apps that are targeted towards testing some aspect of an IP can focus on handling bugs/features in their flows when encountered.

```ruby
class MyIPBlock
  include Origen::Model
  feature :my_feature
end

class MySocBlock < MyIPBlock
  include Origen::Model
end

class MyTop
  include Origen::TopLevel

  def initialize(options = {})
    sub_block :my_block, class_name: 'MySocBlock', inherit: 'MyIPBlock'
  end
end

dut.my_block.has_feature?(:my_feature)
=> true
```

The above works the same for bugs. You can disable bugs and/or feature inheritance by passing either/both `disable_bug_inheritance` or `disable_feature_inheritance` options to the sub_block method:

```ruby
sub_block :my_block, class_name: 'MySocBlock', inherit: 'MyIPBlock', disable_feature_inheritance: true

dut.my_block.has_feature?(:my_feature)
=> false
```

### DIFF
--- a/lib/origen/loader.rb
+++ b/lib/origen/loader.rb
@@ -117,11 +117,23 @@ module Origen
     # If a block definition exists for the given model, then this will load it and apply it to
     # the model.
     # if options[:inherit] is passed, it will first try to load the files for the class name contained
-    # in that option, even if its from a plugin app
+    # in that option, even if its from a plugin app. Additionally, any bugs/features will be inherited
+    # as well unless disable_bug_inheritance or disable_feature_inheritance options are passed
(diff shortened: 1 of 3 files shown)


priyavadan (approved):
Looks good to me, thanks for your contribution!

PR #435, opened 2022-02-11
Add in option to ignore blank lines for the compile method

JamesCreamer:


CodyBriscoe (approved):

priyavadan (approved):
Looks good to me, thanks for the updates!

pderouen:
Seems fine to me too. I guess there is a case where you don't want to ignore blank line diffs?

priyavadan:
> Seems fine to me too. I guess there is a case where you don't want to ignore blank line diffs?

Yup, also just wanted to add in the flexibility without changing default behaviour.

PR #438, opened 2022-04-28
Ruby3 Updates

pderouen:
- Minor updates to get tests passing with Ruby 3
- Web compile is broken again in Windows, disabled that check for Windows GA
- Web compile is also broken with Ruby 3 (Windows & Linux), disabled in GA
- lint crashes when with Ruby 3 in GA, disabled that for now as well
- gem install origen is not able to see anything newer than 0.30, so Ruby 3 GA builds then installs


priyavadan:
@pderouen thanks for the updates. 

I would like to see if that can be fixed before this PR is merged. I  can take a stab at looking into it over the weekend

pderouen:
Thanks for tracking that down. Any idea why lint is failing with Ruby 3? Did I re-break the fix you pushed?

priyavadan:
started a origen_testers branch here with the fix , needs more fixes before it is ruby3 ready

https://github.com/Origen-SDK/origen_testers/tree/feature/ruby_31_wip

priyavadan:
> Thanks for tracking that down. Any idea why lint is failing with Ruby 3? Did I re-break the fix you pushed?

lint is working for me right now... Is it a ruby 3.1 vs ruby 3.0.2p107 thing now?

image

pderouen:
ok, I'll take a closer look at lint with 3.1

pderouen:
> > Thanks for tracking that down. Any idea why lint is failing with Ruby 3? Did I re-break the fix you pushed?
> 
> lint is working for me right now... Is it a ruby 3.1 vs ruby 3.0.2p107 thing now?
> 

It appears to be a 3.0 vs 3.1 issue. Rubocop 0.30.0 crashes for me with Ruby 3.1. It runs fine with 3.0.4. I tried moving to a slightly newer version of rubocop. All of the name spaces changed. I spend a bit of time making updates to the rules only to find that there's not really a 1 to 1 mapping of old rules/settings to new. How were those rule sets created in the first place?

pderouen:
- lint checks re-enabled for the ruby 3 GA checks. The only checks still disabled for Ruby 3 is web compile (requires origen_testers update).
- To re-enable the GA web compile checks for windows, I'll need to debug doc_helpers. It appears to be another occurrence of the issue we saw a while back (GA setup up the workspace on D: and installs origen to C:). Those can be re-enabled in a later PR (same thing we did before).
- To enable Ruby 3.1, we'll need to update the rubocop versioin. That can wait for a future PR in my opinion.
- There is a separate rubygems/ruby3/origen install issue that I am seeing. gem install origen is installing origen 0.34.0. Any attempts to install a newer version fail. My workaround has been to build the origen gem and then install the local .gem file. You can see this in the ruby3 workflow files.

pderouen:
> @pderouen Any performance improvements noticed when running the specs and tests for 3.x?

Seems to be taking about the same amount of time when run with github actions

pderouen:
- Web compile check re-enabled for linux GA workflows
- Filed issues for the windows GA web compile fail [here](https://github.com/Origen-SDK/origen/issues/439) and with [origen_doc_helpers](https://github.com/Origen-SDK/origen_doc_helpers/issues/25)
- Filed an (issue)[https://github.com/Origen-SDK/origen/issues/440] for Rubocop update (to support Ruby 3.1)
- Filed an (issue for gem install origen from ruby 3)[https://github.com/Origen-SDK/origen/issues/441]

Assuming everybody is ok with the changes and the open issues, this PR can be merged. There's not a rush. But, it would be nice to have a version available to use with ruby 3.

ginty:
> It appears to be a 3.0 vs 3.1 issue. Rubocop 0.30.0 crashes for me with Ruby 3.1. It runs fine with 3.0.4. I tried moving to a slightly newer version of rubocop. All of the name spaces changed. I spend a bit of time making updates to the rules only to find that there's not really a 1 to 1 mapping of old rules/settings to new. How were those rule sets created in the first place?

The original rule sets were the defaults that Rubocop provided at the time.
I think I might have changed one or two because they were going against common idioms being used in our code base at the time.
I would recommend that you replace these with the new set of Rubocop defaults and then see what happens when you run it on our code base, maybe you'll be happy enough with any changes or it might drive you to tweak a couple of the rules.

ginty:
Should the tests not also be run on Ruby 2, or is the intention to only support R3 from here?

pderouen:
> Should the tests not also be run on Ruby 2, or is the intention to only support R3 from here?

The tests for Ruby 2.5, 2.6, and 2.7 still run. There are 4 work flows now because of the various issues (raised above):

1. <code>gem install origen</code> with Ruby 3 is installing an ancient version (leads to errors during boot). So, for now there are "_Ruby3" versions of the work flows to build the gem and install locally (which wasn't working right with Ruby 2 and the bundler version).
2. The web compile checks fail under Windows GA. So, there are separate windows work flows that skip the web compile check.

pderouen:
@ginty I think I'll be able to use just the 1 workflow by adding if conditionals. That should make it less convoluted to figure out what's going on with the test setups.

pderouen:
ok, I like this better. Just 1 workflow and it's easy to see what's being run for which conditions.

ginty (approved):
LGTM

coreyeng (changes requested):
What is the TL;DR on the 3.1 support? We have 3.1.2 on the network here but no 3.0. But I only see regressions run on 3.0. Can we add a 3.1 to it?

pderouen:
> What is the TL;DR on the 3.1 support? We have 3.1.2 on the network here but no 3.0. But I only see regressions run on 3.0. Can we add a 3.1 to it?

The rubocop version (0.30.0) is not compatible with Ruby 3.1:

~~~
wrong number of arguments (given 2, expected 1)
C:/Ruby31-x64/lib/ruby/3.1.0/psych.rb:323:in `safe_load'
C:/Users/nxa18793/.origen/gems/ruby/3.1.0/gems/rubocop-0.30.0/lib/rubocop/config_loader.rb:129:in `yaml_safe_load'
~~~

That's the only issue with 3.1 (I think). I can add 3.1 to the checks and have it skip lint.

pderouen:
@coreyeng I can take a look at updating rubocop over the next week or so. I spent a few hours on it when I first ran into it and discovered that we just need all new rule sets if moving to a newer version.

coreyeng:
Sounds good. IMO, its not really _Ruby 3_ ready if we need to axe one of the main commands and checks to make it work.

I can probably help out too towards the end of the month if its still a lingering problem.

pderouen:
@coreyeng I've been sifting through the new lint violations (>6000 with the new default rules) and refining the rules to be closer to what we had before. It's looking like there will be quite a few code changes for lint (100's). I would like to go ahead and start a separate branch just for the lint upgrade. I will get the new branch pushed up to the repo in the next few days.

coreyeng (approved):

coreyeng:
Sounds good!

Issue #440, opened 2022-05-11
Rubocop version requires update to support Ruby 3.1


pderouen:
0.30.0 generates run time errors with Ruby 3.1. Extensive rule set updates will be needed in order to move to a newer revision.

PR #442, opened 2022-05-14
Lint update

pderouen:
Resolves https://github.com/Origen-SDK/origen/issues/440


lib/origen/application/lsf_manager.rb
```diff
@@ -209,7 +209,7 @@ def clear(options)
           if options[:type] == :all
             File.delete(remote_jobs_file) if File.exist?(remote_jobs_file)
             @remote_jobs = {}
-            return
```
pderouen, line 212: This diff is caused by [this](https://github.com/Origen-SDK/origen/blob/d1143f82d2019202b06fa3bec2e6a5b0dc102878/config/rubocop/strict.yml#L4594) rule. Rubocop doesn't indicate that the rule has been updated.

We have this rule [enabled](https://github.com/Origen-SDK/origen/blob/e531b9c03573a6417a4588ca520bcafa8aa41eb8/config/rubocop/strict_enabled.yml#L392-L395) in our existing rules.

Most of the diffs have a similar story.

coreyeng (approved):
Looks good and thanks for doing this! We'll have to bite the bullet and update the linter at some point so probably sooner rather than later is better anyway.

pderouen:
> Looks good and thanks for doing this! We'll have to bite the bullet and update the linter at some point so probably sooner rather than later is better anyway.

Thanks. I'll give @priyavadan and @info-rchitect  a few more days to chime in if they'd like any changes.

info-rchitect (approved):
LGTM, agreed that we need to bite the bullet

PR #443, opened 2022-05-26
Add net smtp runtime dependency

pderouen:
Without this the mailer utility has a runtime error at the <code>require 'net/smtp'</code> on line 1. This happens with Ruby 3.1 only. I didn't stumble across it until I tried to release origen.

### DIFF
--- a/.github/workflows/regression_tests.yml
+++ b/.github/workflows/regression_tests.yml
@@ -4,12 +4,13 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


pderouen:
don't know what the deal is with the Ruby 2.7 checks. It gets stuck in an infinite loop with <code>origen -v</code> complaining about not finding timeout-0.3.0. But, it shows that it got installed. This isn't happening in a local 2.7 install.

pderouen:
might take a while to sort this out

PR #444, opened 2022-05-28
Net smtp

pderouen:
- In Ruby 3.1 the <code>require 'net/smtp'</code> on the first line of the email utility causes an error unless a run time dependency is added for 'net-smtp'
- Unfortunately the introduction of the dependency exposed what appears to be a bug running origen on the GA machine (only in Ruby 2.7). I put in a work around for that bug (see the workflow diffs). I wasn't able to get a workaround for windows with Ruby 2.7. So, I put back in a separate windows workflow skipping Ruby 2.7.
- The workaround cause the Ruby 2.5 checks to fail. So, disabled those as well.
- Also updated boot/app.rb to resolve the bundler deprecation warning. This was an update during debug. But, decided to keep it.
- Put in a fix for a rename file error happening on Windows


pderouen:
@priyavadan @coreyeng @info-rchitect @chrisnappi @redxeth @ginty any thoughts on this?

coreyeng:
We're moving to 2.7 here and the apps I've been releasing are requiring that version. I'm not a big fan of just disabling 2.7 GA checks. From what I can tell though, it looks like Bundler is caught in a loop. May need a new bundler version than what's being used, or something like that?

pderouen:
> We're moving to 2.7 here and the apps I've been releasing are requiring that version. I'm not a big fan of just disabling 2.7 GA checks. From what I can tell though, it looks like Bundler is caught in a loop. May need a new bundler version than what's being used, or something like that?

linux 2.7 checks are still [running](https://github.com/Origen-SDK/origen/actions/runs/2430489979) albeit with the workaround. I tried 1 other bundler version (and several other things). I checked with the 2.7 install on our network (same boot code, ruby version, and bundler version) and the infinite loop isn't happening.

pderouen:
Cancelling this PR. If you want to be able to use Ruby >= 3.1, <code>origen rc tag</code> will error out at the send release email step. The fix is to add a run time dependency to <code>net-smtp</code>, or the mailer code can be skipped based on Ruby version.

PR #445, opened 2023-02-21
Variable max lsf jobs

CodyBriscoe:
Allow ability to modify the max job count before origen lsf handler sleeps to submit jobs

Edit: Update - Added ability to only check the current queue rather than all submitted jobs, as we have a service account that submits lots of jobs, and werent able to submit any origen jobs as there were a few thousand running in a different queue


priyavadan:
these updates should be okay. Thank you!

pderouen (approved):
LGTM!

PR #448, opened 2023-03-04
ability to for a custom rubocop config and support force autocorrect

CodyBriscoe:
when upreving to later origen, there are some lint rules that would be hard to get userbase to follow/would be a hassle for the existing codebase to move to. added option to specify a custom yml instead of the provided one

### DIFF
--- a/lib/origen/commands/lint.rb
+++ b/lib/origen/commands/lint.rb
@@ -18,9 +18,11 @@ http://origen.freescale.net/origen/latest/guides/utilities/lint/
   END
   opts.on('-c', '--correct', 'Correct errors automatically where possible') { options[:correct] = true }
   opts.on('-n', '--no-correct', "Don't correct errors automatically (override if the app default is set to auto correct)") { options[:no_correct] = true }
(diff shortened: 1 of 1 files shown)


info-rchitect (approved):
LGTM

pderouen:
@CodyBriscoe looks good to me as well. Would you recommend different default rulesets? Or perhaps a third option (other than easy or strict)?

CodyBriscoe:
@pderouen I wouldn't recommend different default rulesets, for the most part the current ones are fine there were just a few checks that altered our code functionality (Lint/DisjunctiveAssignmentInConstructor for example) or there were checks that we really should have as they're good checks (Lint/ShadowedArgument) but unfortunately would break a lot of code as well (we're in need of a refactor :D).

If I ever revisit this, what I'd like to do is instead of pointing to a different file to use entirely, instead have an override file where you still point to the default strict or easy ruleset (and therefore pickup any updates when origen updates) and only takes the few rules from your file and applies it on top of the default file with priority.

priyavadan:
Looks good to me, will merge. Thanks for the updated @CodyBriscoe

PR #449, opened 2023-04-12
Enforcing this rule would lead to lots of rework

pderouen:
Applications generated by Origen with the new structure fail this rule and auto correct breaks applications. Probably best to just turn it off.


chrisnappi:
LGTM, thanks for looking out.

redxeth:
@pderouen Paul-- were you going to do a new release?

PR #450, opened 2023-10-07
Catch flush issues when a logger is closed

coreyeng:
If using multiple shells in the same workspace, there's chance logger files will step on each other, particularly `last`, and particularly noticeable with simulations. This will generate an exception, but doesn't affect a single pattern. However, this will kill a `flow` simulation prematurely if it occurs.

This is a quick fix to just catch any errors flushing to a log file and print them to the user, without propagating the exception.

For example, instead of seeing this:

image

You'll see:

image

Note the `warnings` here - just saying stale log file, which is expected, but now results in the warning instead of an exception.

### DIFF
--- a/lib/origen/log.rb
+++ b/lib/origen/log.rb
@@ -202,14 +202,12 @@ module Origen
     # Force the logger to write any buffered output to the log files
     def flush
       @open_logs.each do |logger, file|
-        begin
-          file.flush
-        rescue => e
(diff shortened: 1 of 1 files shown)


pderouen (approved):
Looks good to me

coreyeng:
Released 0.60.12 with this merged.

PR #452, opened 2024-04-11
Allow remotes to disable version prefix; git branch comparator corrected

CodyBriscoe:
Ran into a scenario where I needed to use a remote from a repo controlled by another team. They tag their repo as '1.2.3' not 'v1.2.3', so this update lets remotes disable that prefixing.

Additionally ran into an error where `version == 'HEAD'` was using the VersionString#equal? instead of String#equal?,
which it would fail because it didnt know how to compare the version of 'HEAD', so I forced the string literal comparison with a .to_s call

### DIFF
--- a/lib/origen/remote_manager.rb
+++ b/lib/origen/remote_manager.rb
@@ -139,13 +139,13 @@ module Origen
           if File.exist?("#{dir}/.initial_populate_successful")
             FileUtils.rm_f(version_file) if File.exist?(version_file)
             rc = RevisionControl.new remote: rc_url, local: dir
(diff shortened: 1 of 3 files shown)


CodyBriscoe:
Not sure what's going on with the regressions, having bundler install issues

pderouen:
Thanks for the update.

pderouen (approved):

rlaj (approved):
LGTM

ThaoHuynhFsl:
Ran couple of active apps with this PR and it works fine.  No issue from me

PR #459, opened 2025-03-07
Remote block inheritance improvement and lazy block double on_create fix

CodyBriscoe:
For remote inheritance there was a couple issues:
* remote blocks' sub blocks ended up getting double instantiated. This was at the time handled via override feature that would throw away the block and replace it later, but this was causing issues when the instantiation was editing data structures external to the block itself.
    * Now when the remote block files are inherited, it first delays handling the remote sub_blocks.rb file until after the current block is fully booted. Then, when parsing the remote sub_blocks.rb, any sub_block that already exists for the current block is skipped.
* bugs and features weren't correctly propagating with lazy loading if those bugs/features were sourced from the remote blocks ancestors
    * remote inheritance now sources the remote blocks ancestry chain to collect all features and bugs. Priority is given to the closest ancestor if a conflict were to occur.

Additionally, lazily loaded blocks that were instantiated after the typical on_create hook call would end up calling on_create twice (once via register_callback and again via model_initializer).
* Now if either of those points call on_create, the object will set an instance variable to signal that it has been called and skip the call on the next encounter.

### DIFF
--- a/lib/origen/callbacks.rb
+++ b/lib/origen/callbacks.rb
@@ -12,7 +12,12 @@ module Origen
       # If this object has been instantiated after on_create has already been called,
       # then invoke it now
       if Origen.app.on_create_called?
-        on_create if respond_to?(:on_create)
+        if respond_to?(:on_create)
+          unless @_on_create_called
+            @_on_create_called = true
+            on_create
+          end
+        end
       end
     end
   end
--- a/lib/origen/loader.rb
+++ b/lib/origen/loader.rb
@@ -124,16 +124,6 @@ module Origen
       model = model.model # Ensure we have a handle on the model and not its controller
       loaded = nil
 
-      if options[:inherit]
-        # pass down any bugs/features from the inherited block class
-        unless options[:disable_bug_inheritance]
-          model.class.instance_variable_set(:@bugs, options[:inherit].constantize.bugs.merge(model.class.bugs))
-        end
-        unless options[:disable_feature_inheritance]
-          model.class.instance_variable_set(:@features, options[:inherit].constantize.features.merge(model.class.features))
-        end
-      end
-
       if local_app = options[:app] || model.app
         if options[:path]
           local_full_paths = Array(options[:path])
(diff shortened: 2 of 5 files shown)


JamesCreamer (approved):

coreyeng:
I can't see that the examples cover it either since I can't find any that are using a subblock. Most are timing and pins.

CodyBriscoe:
Hey @coreyeng, can you elaborate on "unloading the reinstantiation when the DUT is reloaded"? My origen knowledge wrt to pattern generation is limited 😅
Could you point to or write (pseudo is fine) an example of when this process occurs?

Is this [load count callback spec](https://github.com/Origen-SDK/origen/blob/d5a192f56120c235b8b13aef9d7eeef1d0a61285/spec/callbacks_spec.rb#L135) an example of the back to back pattern generation you noted: "For example, when running two patterns back-to-back?"

coreyeng:
Correct, but we'd rarely run the same pattern twice. For us, its more in either a simulation flow where we alter the state of the part, reset, the test the state. Or, for generating a very similar pattern with with just a different instance of a block (e.g., adc1 vs adc2), we call Pattern.create() in a loop. Same process, but a DUT gets reinitialized. This might looks like a "origen g adc_test" where its got:

~~~ruby
[0, 1].each do |adc_num|
 Pattern.create(pat_postfix: adc_num) do
  dut.send("adc#{adc_num}".to_sym).test!
 end
end
~~~

This would generate two patterns, adc_test_0, adc_test_1, where we rely on a fresh DUT each time.

Happens very rarely, and hasn't happened in recent memory, but optimizations end up caching the DUT state somewhere and it doesn't get reset.

To test this specific PR, probably an example running generating two patterns in the same process, where each pattern writes the initial value of a block, changes it, then writes the new value, would be the simplest, and would get additional "DUT state is reset" tested by proxy.

For this too, I pulled it down and tried it out and its fine. Sorry for the delay on that.

coreyeng (approved):

PR #460, opened 2025-04-26
locking json gem to 2.10.x as 2.11.x introduced a breaking change to JSON.pretty_generate

CodyBriscoe:
json 2.11 was released earlier this week which is causing the failure, I dug a bit into the registers to_json call but wasn't able to figure it out just yet so locking it down for now.


coreyeng:
Which 2.11 did you try with? Looks like there's been some pushes since the original 2.11.0 and `pretty_generate`: https://github.com/ruby/json/blob/master/CHANGES.md

CodyBriscoe:
@coreyeng I tried with each 2.11 version (0,1,2,3) in the hopes one would work. The latest releases appear slightly newer than they are, as the gems were published but the changelog or github release system or something only got updated recently

CodyBriscoe:
I tried again with the recently released 2.12.0, same issue

PR #463, opened 2026-02-21
Add Error Assistant LLM integration

priyavadan:
Origen::ErrorAssistant
  applies mode precedence for off/forced/opt_in
  falls back display name when friendly name missing
[WARNING]    0.070[0.070]    || Error Assistant timed out after 0.01s
  times out best effort and returns nil
  passes site config settings into plugin context

```
[Error Assistant Response]
Based on the stack trace provided, here's my diagnosis:

## 1) Most likely root cause
This appears to be a test failure in the Error Assistant functionality within the Origen application's test suite. The exception is occurring at line 147 of the error_assistant_spec.rb file during RSpec test execution, suggesting either a test assertion failure or an unhandled exception within the Error Assistant feature being tested.

## 2) Suggested fix steps
• Check the specific test at line 147 in `spec/error_assistant_spec.rb` to understand what functionality is being tested
• Run the test with verbose output (`rspec --format documentation`) to see the actual error message and test description
• Verify that any dependencies or mock objects required by the Error Assistant are properly set up in the
  live test: raises and requests a real solution (manual)
```

Finished in 7.41 seconds (files took 0.39531 seconds to load)
5 examples, 0 failures

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ gem "origen_sim", git: "https://github.com/Origen-SDK/origen_sim.git"
 
 # Required to run the concurrent test case patterns from OrigenSim
 gem 'origen_jtag'
+gem 'origen_llm', "~>0.1.2"
 
 # Pull in the latest and greatest app generator templates so that they can be
 # packaged into Origen releases
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ gem "origen_sim", git: "https://github.com/Origen-SDK/origen_sim.git"
 
 # Required to run the concurrent test case patterns from OrigenSim
 gem 'origen_jtag'
-gem 'origen_llm', "~>0.1.2"
+gem 'origen_llm', "~>0.1.4"
 
 # Pull in the latest and greatest app generator templates so that they can be
 # packaged into Origen releases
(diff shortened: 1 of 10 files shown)


Gemfile
```diff
@@ -21,6 +21,7 @@ gem "origen_sim", git: "https://github.com/Origen-SDK/origen_sim.git"
 
 # Required to run the concurrent test case patterns from OrigenSim
 gem 'origen_jtag'
+gem 'origen_llm', "~>0.1.2"
```
coreyeng: As a general question, is it worth adding `origen_llm` to the gemspec for application usage too? Or is this intended for just the core at this time?

priyavadan: fair call, good catch. It should be in the gemspec. I've updated it.

coreyeng (approved):

pderouen:
This is a nice feature! Any web docs update needed?

priyavadan:
> This is a nice feature! Any web docs update needed?

I tried to sneak it in @pderouen ... yup ok, will add some docs

pderouen (approved):
love it!

ThaoHuynhFsl (approved):
looking forward to it

PR #466, opened 2026-08-09
Skip version check if no bug exists

ikokjyan:
Bug observed when using `dut.has_bug?` with a dut model that does not have the `@version` attribute


priyavadan (approved):
LGTM

JamesCreamer (approved):

